Issue #32, opened 2020-08-31
Specify optional ID to "Operator::AddTerm()"


hoppe93:
This could allow us to easily identify different ``EquationTerm``s across the code and pick out contributions from individual terms in, for example, the ``OtherQuantityHandler``.

hoppe93:
@vaoskar I have implemented a new way of accessing individual ``EquationTerm``s now and pushed to the branch iss-32-42. Once we have fixed the bugs currently preventing the examples from running, would you mind having a look and verifying that I implemented the other quantity ``fluid/radiation`` correctly?

Embreus:
Fixed by #53.

PR #43, opened 2020-09-11
Spi

vaoskar:
Request for merging implementation of compton source and ability to store contribution to SetVectorElements from individual terms into master. Ability to choose which of the avalanche formulas implemented in RunawayFluid that should be used in the simulation is also implemented via an alternative avalanche term AvalancheGrowthTermHesslow. A bug when setting the IdentityTerm in j_re when using fluid runaways was also corrected.


src/Settings/Equations/n_re.cpp
```diff
@@ -57,13 +59,22 @@ void SimulationGenerator::ConstructEquation_n_re(
...
         // XXX: assume same momentum grid at all radii
         real_t pMax = hottailGrid->GetMomentumGrid(0)->GetP1_f(hottailGrid->GetNp1(0));
         Op_nRE->AddTerm(new AvalancheSourceRP(fluidGrid, eqsys->GetUnknownHandler(),pMax, -1.0, AvalancheSourceRP::RP_SOURCE_MODE_FLUID) );
     }
+/*
+AvalancheSourceRP::AvalancheSourceRP(
+    FVM::Grid *kineticGrid, FVM::UnknownQuantityHandler *u,
+    real_t pCutoff, real_t pMin, RPSourceMode sm
+)
+*/
```
Embreus, line 77: what is this comment for?

src/Settings/Equations/j_re.cpp
```diff
@@ -60,7 +60,7 @@ void SimulationGenerator::ConstructEquation_j_re(
     // Otherwise, we set it to zero...
     } else {
         eqn->AddTerm(new FVM::IdentityTerm(fluidGrid, Constants::ec * Constants::c));
-        eqsys->SetOperator(id_j_re, id_j_re, eqn, "j_re = e*c*n_re");
+        eqsys->SetOperator(id_j_re, id_n_re, eqn, "j_re = e*c*n_re");
```
Embreus, line 63: Nice catch! Sorry about that, very sloppy on my part.

src/Equations/Fluid/ComptonRateTerm.cpp
```diff
@@ -0,0 +1,109 @@
...
+        const len_t np1 = this->grid->GetMomentumGrid(ir)->GetNp1();
+        const len_t np2 = this->grid->GetMomentumGrid(ir)->GetNp2();
+
+        // Insert at p=0, xi=1
+        vec[offset + np1*(np2-1) + 0] += scaleFactor*gamma[ir];
+
+        offset += np1*np2;
+    }
+}
+
```
Embreus, line 109: I would suggest making this a term that should be added in the (id_n_re, id_n_tot) block of the eqsys in n_re.cpp. Then you get some things for free which helps make setvectorelements, jacobian etc cleaner (and lets us keep matrixelements implicit too and not use the rhs only). If you don’t feel comfortable doing it, and don’t mind, I could look into it after the merge - it’s a small change really.
We should also consider adding pstar derivatives. In the avalanche term we have an approximate expression where we assume 
pcrit ~ 1/sqrt(ntot*(E-Eceff)).
Could perhaps make that approximation here as well?

include/DREAM/Equations/Fluid/ComptonRateTerm.hpp
```diff
@@ -0,0 +1,42 @@
...
+        ComptonRateTerm(FVM::Grid* g, FVM::UnknownQuantityHandler *u, 
+                RunawayFluid *ref, real_t scaleFactor = 1.0);
+        ~ComptonRateTerm();
+
+        void AllocateGamma();
+        void DeallocateGamma();
+        
+        virtual bool GridRebuilt() override;
+        virtual len_t GetNumberOfNonZerosPerRow() const { return 1; }
+        virtual len_t GetNumberOfNonZerosPerRow_jac() const { return 1; }   /* XXX TODO XXX */
```
Embreus, line 33: 1 is a good start (for ntot). Remember to add 1 when we include E derivatives later.

include/FVM/Equation/Operator.hpp
```diff
@@ -126,6 +128,15 @@ namespace DREAM::FVM {
...
         void SetMatrixElements(Matrix*, real_t*);
         void SetVectorElements(real_t*, const real_t*);
+
+        /**
+         * Returns the contribution from the SetVectorElements-function from an individual term
+         * To be used for accessing contributions from individual terms 
+         * so they can be saved to the output as "other quantities"
+         * TODO: come up with a better way to handle id's for individual terms, 
+         * now one has to check in which order the terms are added to the operator!
+         */
```
Embreus, line 138: This is fine for now, we’ve got an issue (#32) to sort this out

include/DREAM/Equations/Fluid/ComptonRateTerm.hpp
```diff
@@ -0,0 +1,42 @@
...
+#define _DREAM_EQUATION_FLUID_COMPTON_RATE_TERM_HPP
+
+#include "DREAM/Equations/RunawayFluid.hpp"
+#include "FVM/Equation/DiagonalComplexTerm.hpp"
+//#include "FVM/Equation/EquationTerm.hpp"
+#include "FVM/Grid/Grid.hpp"
+#include "FVM/Matrix.hpp"
+#include "FVM/UnknownQuantityHandler.hpp"
+/**
+ * Implementation of a class which represents the Gamma_tritium contribution to the n_re equation.
```
Embreus, line 11: Gamma_tritium? :P

py/DREAM/Settings/Equations/RunawayElectrons.py
```diff
@@ -61,6 +67,13 @@ def setDreicer(self, dreicer):
         """
         self.dreicer = dreicer
 
+    def setCompton(self, compton):
+        """
+        Specifies which model to use for calculating the
+        compton runaway rate.
+        """
+        self.compton = compton
+
```
Embreus, line 76: Down the line we should be able to set the photon flux here. 	COMPTON_RATE_ITER_DMS could still be used, as a shortcut for 1e18.

Embreus (commented):
I’m very impressed! Really solid work. 
Giving my first thoughts here; first time using this feature.

src/CMakeLists.txt
```diff
@@ -65,6 +66,7 @@ set(dream_equations
 set(dream_equation_headers
     "${PROJECT_SOURCE_DIR}/include/DREAM/Equations/Fluid/CurrentFromConductivityTerm.hpp"
     "${PROJECT_SOURCE_DIR}/include/DREAM/Equations/Fluid/AvalancheGrowthTerm.hpp"
+    "${PROJECT_SOURCE_DIR}/include/DREAM/Equations/Fluid/AvalancheGrowthTermHesslow.hpp"
```
Embreus, line 69: this file is not commited, right? 

I am thinking that we may not want to create a separate class for using the other formula, but rather giving a flag to the avalanche growth term and let that decide which Gamma to use in the calculation.

Embreus: and then also pass that flag to RunawayFluid so that only the formula that is used is evaluated.

py/DREAM/Settings/OtherQuantities.py
```diff
@@ -28,9 +28,12 @@ class OtherQuantities:
         'fluid/Eceff',
         'fluid/GammaAva', 'fluid/GammaAvaAlt',
         'fluid/pCrit',
+        'fluid/GammaCompton',
+        'fluid/GammaDreicer',
```
hoppe93, line 32: I would prefer if we could keep this list in alphabetical order (I realize that the lack of order most likely predates your changes Oskar, sorry about that :P)

src/Equations/Fluid/ComptonRateTerm.cpp
```diff
@@ -0,0 +1,109 @@
...
+ */
+void ComptonRateTerm::SetJacobianBlock(const len_t, const len_t derivId, FVM::Matrix *jac, const real_t*){
+    const len_t nr  = this->grid->GetNr();
+    len_t offset = 0;
+
+    for(len_t ir=0;ir<nr;ir++){
+        const len_t np1 = this->grid->GetMomentumGrid(ir)->GetNp1();
+        const len_t np2 = this->grid->GetMomentumGrid(ir)->GetNp2();
+
+        if(derivId==id_n_tot){
```
hoppe93, line 75: It's probably better to put this if statement outside of the for loop for now. This saves a few branch checks (which should generally be avoided within loops)

src/Equations/Fluid/ComptonRateTerm.cpp
```diff
@@ -0,0 +1,109 @@
...
+            jac->SetElement(offset+np1*(np2-1)+0,offset+np1*(np2-1)+0,this->gamma[ir]/this->data_n_tot[ir]);
```
hoppe93, line 76: I assume that this has been benchmarked and doesn't need any further volume elements...

src/Equations/RunawayFluid.cpp
```diff
@@ -866,6 +866,28 @@ real_t* RunawayFluid::evaluatePartialContributionAvalancheGrowthRate(len_t deriv
...
+ */
+real_t* RunawayFluid::evaluatePartialContributionAvalancheGrowthRateAlt(len_t derivId) {
+    real_t *dGamma = new real_t[nr];
+    if( !( (derivId==id_Eterm) || (derivId==id_ntot) ) )
+        for(len_t ir = 0; ir<nr; ir++)
+            dGamma[ir] = 0;
+
+    // set dGamma to d(Gamma)/d(E_term)
+    for(len_t ir=0; ir<nr; ir++)
+        dGamma[ir] = avalancheGrowthRateAlt[ir] / ( Eterm[ir] - effectiveCriticalField[ir] );
```
hoppe93, line 880: Is there a ``if (Eterm[ir] > effectiveCriticalField[ir]) ...`` missing here? If this method is called even if E is too low, then this would just output nonsense? And if E = Eceff DREAM would crash as it currently stands.

src/Equations/RunawayFluid.cpp
```diff
@@ -866,6 +866,28 @@ real_t* RunawayFluid::evaluatePartialContributionAvalancheGrowthRate(len_t deriv
     return dGamma;
 }
 
+/**
+ * Same as the function above, but for the formula published by Hesslow et al NF 2019
```
hoppe93, line 870: Instead of ``function above``, you could perhaps give the functions name explicitly? In case code moves around in the future

src/Equations/RunawayFluid.cpp
```diff
@@ -866,6 +866,28 @@ real_t* RunawayFluid::evaluatePartialContributionAvalancheGrowthRate(len_t deriv
...
+ */
+real_t* RunawayFluid::evaluatePartialContributionAvalancheGrowthRateAlt(len_t derivId) {
+    real_t *dGamma = new real_t[nr];
+    if( !( (derivId==id_Eterm) || (derivId==id_ntot) ) )
+        for(len_t ir = 0; ir<nr; ir++)
+            dGamma[ir] = 0;
```
hoppe93, line 876: What is the purpose of this code? It seems it never has any effect, since ``dGamma`` is always overwritten on line 880.

src/Equations/RunawayFluid.cpp
```diff
@@ -866,6 +866,28 @@ real_t* RunawayFluid::evaluatePartialContributionAvalancheGrowthRate(len_t deriv
     return dGamma;
 }
 
+/**
+ * Same as the function above, but for the formula published by Hesslow et al NF 2019
+ */
+real_t* RunawayFluid::evaluatePartialContributionAvalancheGrowthRateAlt(len_t derivId) {
+    real_t *dGamma = new real_t[nr];
```
hoppe93, line 873: This is more of a comment to @Embreus, I guess, but I think we should try to **not** allocate memory in these functions. Because I suspect that these are called on every jacobian rebuild? In which case we include a bunch of expensive memory allocations, which also risk causing memory leaks.

src/OtherQuantityHandler.cpp
```diff
@@ -38,12 +42,17 @@ using namespace std;
  */
 OtherQuantityHandler::OtherQuantityHandler(
     CollisionQuantityHandler *cqtyHottail, CollisionQuantityHandler *cqtyRunaway,
-    PostProcessor *postProcessor, RunawayFluid *REFluid,
+    PostProcessor *postProcessor, RunawayFluid *REFluid, FVM::UnknownQuantityHandler *unknowns, std::vector<UnknownQuantityEquation*> *unknown_equations, Settings *s,
```
hoppe93, line 45: Remove ``Settings`` (see my comment for ``OtherQuantityHandler.hpp``)

src/OtherQuantityHandler.cpp
```diff
@@ -177,10 +186,12 @@ void OtherQuantityHandler::DefineQuantities() {
     const len_t n1_re = (this->runawayGrid==nullptr ? 0 : this->runawayGrid->GetMomentumGrid(0)->GetNp1());
     const len_t n2_re = (this->runawayGrid==nullptr ? 0 : this->runawayGrid->GetMomentumGrid(0)->GetNp2());
 
+    const real_t *x = unknowns->GetUnknownData(id_ncold);
+
     // HELPER MACROS (to make definitions more compact)
     // Define on fluid grid
     #define DEF_FL(NAME, DESC, FUNC) \
-        this->all_quantities.push_back(new OtherQuantity((NAME), (DESC), fluidGrid, 1, FVM::FLUXGRIDTYPE_DISTRIBUTION, [this](QuantityData *qd) {FUNC}));
+        this->all_quantities.push_back(new OtherQuantity((NAME), (DESC), fluidGrid, 1, FVM::FLUXGRIDTYPE_DISTRIBUTION, [this,x](QuantityData *qd) {FUNC}));
```
hoppe93, line 194: Prefer ``&x`` to provide a reference to ``x`` rather than copying the address it's pointing to into a new variable

src/OtherQuantityHandler.cpp
```diff
@@ -177,10 +186,12 @@ void OtherQuantityHandler::DefineQuantities() {
     const len_t n1_re = (this->runawayGrid==nullptr ? 0 : this->runawayGrid->GetMomentumGrid(0)->GetNp1());
     const len_t n2_re = (this->runawayGrid==nullptr ? 0 : this->runawayGrid->GetMomentumGrid(0)->GetNp2());
 
+    const real_t *x = unknowns->GetUnknownData(id_ncold);
```
hoppe93, line 189: ``x`` should be called ``n_cold``

hoppe93 (changes requested):
I have a number of comments which I'd like for you to fix before mergeing. The comments concerning the retrieval of data from individual equation terms could be ignored now as I intend to fix them myself in issue #32.

src/Equations/RunawayFluid.cpp
Embreus: Indeed, I agree that this should be revised.

src/Equations/RunawayFluid.cpp
Embreus: Looks like a logic error (that I introduced). Should be followed by return dGamma after the for loop (inside the if statement)

src/Equations/RunawayFluid.cpp
Embreus: avalancheGrowth.. explicitly (or well, quite implicitly actually, but it’s there) has a factor of (E-Eceff) that this is intended to cancel. This is supposed to be applied for all E. Because of the way the calculation is structured, it is a bit tricky to extract the resulting factor explicitly. At least we should handle the special case E=Eceff somehow (just set to 0? it’s quite unlikely to ever occur)

src/Equations/Fluid/ComptonRateTerm.cpp
Embreus: It should be legit. It’s just a source term, so it doesn’t involve any trickery with FVM but is just a simple evaluation at the cell center.

src/Equations/Fluid/ComptonRateTerm.cpp
vaoskar: I think I know how to do it now, I was just a bit narrow-minded when I started and for some reason wanted to keep all the fluid runaway terms on the same operator. We'll see who gets at it first ;).

fvm/Equation/Operator.cpp
```diff
@@ -14,7 +14,9 @@ using namespace std;
 /**
  * Constructor.
  */
-Operator::Operator(Grid *grid) : grid(grid) {}
+Operator::Operator(Grid *grid) : grid(grid) {
+vectorElementsSingleTerm=new real_t(this->grid->GetNCells());
```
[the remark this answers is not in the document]
vaoskar: I have also thought about point 3, but since the macros in OtherQuantityHandle.cpp want functions that return the data to be stored instead of void-functions with an output array as an input parameter, I implemented it this way to be compatible with that (although that does, of course, not mean that it is the best way to do it).

include/DREAM/OtherQuantityHandler.hpp
```diff
@@ -21,11 +25,21 @@ namespace DREAM {
...
+        len_t id_ncold;
+        UnknownQuantityEquation *eqn_Tcold;
+        len_t id_term_rad;
+        
+
     public:
         OtherQuantityHandler(
             CollisionQuantityHandler*, CollisionQuantityHandler*,
-            PostProcessor*, RunawayFluid*,
+            PostProcessor*, RunawayFluid*, FVM::UnknownQuantityHandler*, std::vector<UnknownQuantityEquation*>*, Settings*,
```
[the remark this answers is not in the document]
vaoskar: I thought it convenient to not have to add more input arguments whenever one wants to add another term to the output, so this "error" is actually deliberate. Is there a good way to have access to all information about which terms are actually included in the OtherQuantityHandler or is adding more input arguments the best solution anyway?

src/Equations/Fluid/ComptonRateTerm.cpp
vaoskar: Only benchmarked in fluid mode... You mentioned some similar fix in the Dreicer source a while ago, I can have a look at that.

src/Equations/RunawayFluid.cpp
vaoskar: Since we allow the avalanche growthrate to turn negative (to emulate dissipation), I don't see the issue with E<Eceff, but E=Eceff would certainly cause problems

include/DREAM/OtherQuantityHandler.hpp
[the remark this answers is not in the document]
hoppe93: I actually favour adding more input arguments. Alternatively, we could add an enum to the ``OtherQuantityHandler`` containing all the options we need to store. Just inserting a ``Settings`` object is not very verbose (it's unclear how it will actually be used), and it's not how it was intended to be used.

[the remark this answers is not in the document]
hoppe93: (another related reason is that we should try to limit the amount of hard-coding of settings names. By keeping the ``Settings`` object to the ``SimulationGenerator`` we can keep all that hard-coding to files under ``src/Settings/``)

src/Equations/Fluid/ComptonRateTerm.cpp
Embreus: I was also thinking that we shouldn't allocate memory for gamma and store in in the equation term, since it's already stored in runawayfluid and can be retrieved when it is needed in the Set... functions.

vaoskar: Regarding pstar derivatives: should we use pstar or the same pcrit that appears in the avalanche term in the compton source?

Embreus: What is your opinion? 
I want to try it out with a benchmark against a kinetic description of the compton source, and then pick the most accurate option.

examples/compton/plot_Ebalance.py
```diff
@@ -0,0 +1,187 @@
...
+ds.other.include('fluid', 'lnLambda','nu_s','nu_D')
+
+
+# Save settings to HDF5 file
+ds.save('init_settings.h5')
+runiface(ds, 'output_init.h5', quiet=False)
+#### Ionization #############
+ds2 = DREAMSettings(ds)
+
+ds2.fromOutput('output_init.h5')
```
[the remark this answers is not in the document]
vaoskar: If you have called the ds.output.setFilename(…) function before saving the settings object, this line is not needed. As it is now, it is however possible to run the simulation using runiface without doing that (the output filename is instead specified in the call to runiface), and then DREAMSettings(ds) calls fromOutput with the default filename 'output.h5'. Wouldn't it be better if runiface used the filename set to the settings object?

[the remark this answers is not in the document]
hoppe93: Ok, that (kind of) makes sense. The ``runiface`` function was written before the output filename could be specified in the ``DREAMSettings`` object. I agree that there should be more interaction between ``runiface`` and the given settings object, although I would argue that it could still be logical for ``runiface`` *not* to load the filename from the settings object. This is because I think it makes sense to be able to run DREAMi easily without saving the output data to disk. However, it would still make sense for ``runiface`` to update the ``ds.output.filename`` option when the output filename is specified as a parameter to ``runiface``. I think this would allow us to remove line 149.

hoppe93 (approved):
Great! I am very happy with these changes. Well done!

Issue #37, opened 2020-09-09
Set tolerances of the nonlinear solver for individual unknowns


Embreus:
Implement interface support to adjust the absolute and relative tolerances of the non-linear solver of single unknowns. 

The present method solver.SetTolerance(reltol) could be a shortcut to set tolerances of all unknowns, but in the future it would be good to set tolerances of individual unknowns somehow.

Embreus:
Not sure how the interface should look though without being too clumsy?
solver.SetTolerance(reltol_n_re, abstol_n_re, reltol_T_cold, abstol_T_col, ...) 
seems really cumbersome and generalises poorly with new unknowns, but the functionality would be useful.

Let each unknown in the eqsys have its own SetTolerance method?

hoppe93:
Since I think we should be able to set tolerances separately for the ``Solver`` and ``TimeStepper``, I would probably favour an interface similar to the new ``TransportSettings`` interface, i.e. we have a ``ToleranceSettings`` class which is created for ``Solver`` and ``TimeStepper``. The ``ToleranceSettings`` keeps a list of unknowns and their relative and absolute tolerances. By default the list is empty, in which case the kernel applies default tolerances?

Next, to set a tolerance, I imagine we would have something like
```python
ds.solver.tolerances.set('n_re', abstol=1e5, reltol=1e-6)
```
where we specify the name of the unknown as a string. To make things a bit safer, we could add a list of unknown names as constants to (perhaps) ``EquationSystem`` which can be accessed and checked against by the ``ToleranceSettings`` object (to avoid misspellings).

We could then also imagine setting all tolerances in one go with a syntax like
```python
ds.solver.tolerances.set(abstol=1e5, reltol=1e-6)
```

Embreus:
Fixed by hoppe93 in 082969c

PR #49, opened 2020-09-22
Add support for MUMPS and ToleranceSettings

hoppe93:
This branch adds support for using MUMPS as the linear solver (in both the *linearly implicit* and *nonlinear* solvers). This branch also implements a general ``ToleranceSettings`` interface (fixing issue #37) which allows us to set custom relative and absolute tolerances for all unknowns in the ``Solver`` and ``TimeStepper`` (separately). Tolerance checks can also be disabled on selected unknowns, which allows, for example, the adaptive time stepper to only seek good convergence in a particular unknown.

### DIFF
--- a/doc/sphinx/source/compiling.rst
+++ b/doc/sphinx/source/compiling.rst
@@ -45,3 +45,15 @@ where ``NTHREADS`` is the number of CPU threads available on your computer.
 will go).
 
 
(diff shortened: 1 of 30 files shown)


Embreus (approved):
Looks fabulous!

PR #51, opened 2020-09-24
Transport

Embreus:
This branch implements a general transport term that can be used to prescribe arbitrary advection-diffusion transport in the f_hot, f_re, n_re, psi_p and T_cold equations.


Embreus (commented):
Looks great to me now!

py/DREAM/Settings/Equations/PoloidalFlux.py
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        return data
+
+
+    def verifySettings(self):
+        """
+        Verify that the settings of this unknown are correctly set.
+        """
+
```
hoppe93, line 40: A ``pass`` seems to be missing here?

hoppe93 (approved):
Looks good to me now!

PR #53, opened 2020-09-25
Storage of data from individual EquationTerms as OtherQuantity

hoppe93:
This pull request reimplements the ability to store data from individual ``EquationTerm``s as ``OtherQuantity``s. With the new method we avoid allocating extra memory for storage and hardcoding indices into the list of terms in ``Operator``s.


hoppe93:
@vaoskar Could you check that I've implemented ``fluid/radiation`` ``OtherQuantity`` properly? I think I'm taking the data from the correct ``EquationTerm``, but it would be great if you could confirm this.

Embreus:
Using the plot_Ebalance.py script in the compton example folder that Oskar used to illustrate the energy balance, I have verified that the new implementation works as intended.

PR #57, opened 2020-09-30
Kinetic-kinetic boundary condition

hoppe93:
This pull request implements a new boundary condition to be specified on the boundary between the kinetic hot-tail and kinetic runaway grids (and also on fluid runaway grid when the kinetic RE grid is enabled).


Embreus (approved):
Have skimmed through all the lines and can’t see any showstoppers! Champagne?

PR #58, opened 2020-10-01
Integrate SynchrotronTerm in distribution

hoppe93:
Synchrotron losses can now be included in all 3D Fokker-Planck equations.


hoppe93:
Should we implement a synchrotron test first? If so, how should it be designed?

Embreus:
> Should we implement a synchrotron test first? If so, how should it be designed?

A possibility would be to set up a simulation resulting in a bump on tail distribution, and compare the location of the maximum with a CODE simulation.

Embreus (changes requested):
Looks ready to go as soon as the test runs!

hoppe93:
No, this was just a merge with master to get the new xi grids. I still have to make the synchrotron tests run before we merge into master...

Embreus:
> No, this was just a merge with master to get the new xi grids. I still have to make the synchrotron tests run before we merge into master...

Oh, never mind then! Misinterpreted my notification.

PR #63, opened 2020-10-12
Merge development branch

Embreus:
The main feature of this branch is an extensive rework of how n_cold is defined in terms of the distribution when running with COLLFREQMODE_FULL.  

This is achieved by the introduction of:
- a "hot region" momentum cutoff pCutoff (controlled in the interface via HotElectronDistribution.py). This is defined such that "cold" quantities are defined by integrating the distribution function up to p=pCutoff (and "hot" quantities as the remainder, such that the total is always the full momentum integral, as before).
- a new unknown "S_particle", which is the amplitude of a particle source which enforces that the integral of the distribution over the cold region becomes n_cold. This ensures that the density carried by the cold part of the distribution adapts when electrons are created or removed via ionization or transport.

Now, the equations for n_tot and j_tot are always of the form "x_tot = x_cold + x_hot + x_re" regardless of collfreqmode, and only the detailed definitions of the components differ. 

The branch also introduces:
- a significant extension of MomentQuantity to accomodate the hot region cutoff in all integral quantities, as well as additional support to specify the jacobian of the integrand
- the XiBiUniformGridGenerator which allows non-uniform xi grids (piece-wise uniform with two regions of separate resolution)
- kinetic ionization (with preliminary fitted parameters that should be improved upon), which is designed such that the integral over a Maxwellian reproduces the ADAS ionization rates, but is also valid up to ultrarelativistic energies.
- a rework of the hot current equation term in 'hottail' mode, nxi=1 (HotTailCurrentDensityFromDistributionFunction). It is now defined as a smooth integrand rather than a piecewise one, which should improve the convergence behavior.
- accurate temperature jacobian of collision frequencies (replacing the old [collisions] ~ T^{-3/2} approximation)
- (fluid) runaway generation made independent of the sign of electric field; now the dynamics _should_ be independent of sign of the electric field (not yet tested)
- j_re now accounts for the sign of acceleration in the conversion from n_re

Known issues:
- the new hottail mode (nxi=1) is not tested in selfconsistent scenarios.
- the stability of the new COLLFREQMODE_FULL is not thoroughly explored in challenging simulations with self-consistent temperature and ionization: there have been indications that the solver may sometimes not converge in such cases
- kinetic ionization with approximate jacobian does not yet have an n_hot jacobian, nor have we implemented the corresponding ionization heating term in the temperature equation


tests/physics/code_runaway/code_runaway.py
```diff
@@ -130,7 +126,7 @@ def run(args):
     nE, nT = T.shape
     rr     = np.zeros((nE, nT))
     rrFull = np.zeros((nE, nT, nt))
-    for i in range(0, nE):
+    for i in range(1, nE):
```
Embreus, line 129: I am looking over this again. I struggle with reproducing the runaway test in the new branch, and I _suspect_ that in master we only get agreement because the DREAM simulation is underresolved by just the right amount. I'll try to verify this. 

We obviously shouldn't skip the first points like I do here, but perhaps instead look at increasing the tolerance of the test (in case I am not mistaken here).

Embreus: ^---- nope, this was not it. I'll keep looking for an explanation..

Embreus: I think I've figured it out. On the new branch, the bulk is being depleted so that n_cold decreases in the cases where n_re is the largest. Thus, the friction decreases (and hence the effective Dreicer field, as we saw in NORSE) and the runaway rate increases  In master, ncold is kept constant, and this does not happen. I will try to solve it by prescribing ncold instead of evaluating it self-consistently.

Embreus: fixed this in the latest commit

fvm/Equation/MomentQuantity.cpp
```diff
@@ -66,10 +204,38 @@ bool MomentQuantity::GridRebuilt() {
  * x:       Value of the unknown quantity.
  */
 void MomentQuantity::SetJacobianBlock(
-    const len_t unknId, const len_t derivId, Matrix *jac, const real_t* /*x*/
+    const len_t unknId, const len_t derivId, Matrix *jac, const real_t* x
 ) {
-    if (derivId == fId && unknId == fId)
+    if (derivId == unknId)
```
Embreus, line 209: this should be ok right? unknId should in practice always be fId

fvm/Equation/MomentQuantity.cpp
```diff
@@ -4,60 +4,198 @@
...
- * linearly on the unknown to which this moment operator is applied
- * (usually the distribution function f)).
+ * returns the momentum grid spacing dp at momentum p0
+ */
+real_t FindThresholdStep(real_t p0, MomentumGrid *mg){
+    const real_t *p = mg->GetP1();
+    for(len_t i=0; i<mg->GetNp1(); i++)
+        if(p[i]>p0)
+            return mg->GetDp1(i);
+    return p0 - mg->GetP1(mg->GetNp1()-1); // what else to do if p0 is outside the grid?
```
hoppe93, line 79: I don't know what the optimal solution is here, but if it is maybe we should remove the comment?

fvm/Equation/MomentQuantity.cpp
hoppe93: For a given ``MomentQuantity`` *object*, ``unknId`` will only take one value, and that is the ID of the unknown to which the operator is applied. So as long as all ``MomentQuantity``s are applied to ``f``, this should be fine.

hoppe93: (actually, it's probably always fine. I think this is exactly the intended behaviour in every case where the ``MomentQuantity`` is involved)

fvm/UnknownQuantityHandler.cpp
```diff
@@ -131,6 +131,12 @@ real_t UnknownQuantityHandler::GetUnknownDataPreviousTime(const len_t qty) {
 real_t *UnknownQuantityHandler::GetUnknownData(const len_t qty) {
     return unknowns[qty]->GetData();
 }
+/**
+ * name: Name of unknown quantity.
+ */
+real_t *UnknownQuantityHandler::GetUnknownData(const std::string& name) {
+    return GetUnknownData(GetUnknownID(name));
```
hoppe93, line 138: Hmm... Where is this method used? The ``GetUnknownID(...)`` method adds quite a bit of overhead so it should not be used in computationally intensive parts of the code.

Embreus: This replaced various occurences along the lines of

len_t id_x = unknows->GetUnknownId(...)
const real_t *data_x = unknowns->GetUnknownData(id_x)

where id_x wasn’t otherwise used. I don’t believe this was ever used in places with many calls — I’ll review this.

fvm/Equation/MomentQuantity.cpp
Embreus: This is due for a slight rework, as Ida identified a case where the method can lead to undesired behavior.

src/Equations/RunawayFluid.cpp
```diff
@@ -846,7 +847,7 @@ real_t RunawayFluid::evaluateNeoclassicalConductivityCorrection(len_t ir, real_t
     else if(!collisionLess){
         // qR0 is the safety factor multiplied by R0
         const real_t qR0 =  rGrid->GetVpVol(ir)*rGrid->GetVpVol(ir)*rGrid->GetBTorG(ir)*rGrid->GetFSA_1OverR2(ir)*rGrid->GetFSA_NablaR2OverR2(ir)
-             / (4*M_PI*M_PI*Constants::mu0*unknowns->GetUnknownData(unknowns->GetUnknownID(OptionConstants::UQTY_I_P))[ir]);     
+             / (4*M_PI*M_PI*Constants::mu0*unknowns->GetUnknownData(OptionConstants::UQTY_I_P)[ir]);
```
hoppe93, line 850: From what I can tell, this gets called at least ``nr*nMultiples`` times every time the equation system is rebuilt (not counting when used for building the Jacobian). So this is an example of where I think we should prefetch the ID of ``I_p`` and use that when calling ``GetUnknownData()``. As a matter of fact, I changed this on the ``bounceaverage`` branch last night, so could wait until that branch is ready to merge to fix this (but of course merge ``development`` to ``master`` first).

src/Equations/RunawayFluid.cpp
```diff
@@ -817,10 +818,10 @@ real_t RunawayFluid::evaluateSauterElectricConductivity(len_t ir, real_t Zeff){
  * Returns the Braams-Karney electric conductivity of a relativistic plasma.
  */
 real_t RunawayFluid::evaluateBraamsElectricConductivity(len_t ir, real_t Zeff){
-    len_t id_Tcold = unknowns->GetUnknownID(OptionConstants::UQTY_T_COLD);
-    real_t *T_cold = unknowns->GetUnknownData(id_Tcold);
+    if(Zeff<0)
+        throw FVM::FVMException("Conductivity: Negative Zeff provided, aborting.");
+    real_t *T_cold = unknowns->GetUnknownData(OptionConstants::UQTY_T_COLD);
```
hoppe93, line 823: (I also changed this last night on ``bounceaverage``)

fvm/UnknownQuantityHandler.cpp
Embreus: > I don’t believe this was ever used in places with many calls — I’ll review this.

This was clearly not true. I fixed a few problematic uses in the latest commit, but will leave the ones you indicate below for the bounceaverage merge.

hoppe93 (approved):
Looks great! I don't have any further comments!

PR #68, opened 2020-10-19
New heateq

Embreus:
The heat equation has been revised, by changing the definition of the total electron energy from

W = 1.5nT + W_binding

to simply

W = 1.5nT.

This was accomplished by adding the binding-energy contribution -dW_binding/dt due to ionization and recombination into the RadiatedPowerTerm. A few minor bugs in the basicE example were also fixed.

The changes have been tested by verifying that the output of basicE (which simulates a self-consistent temperature drop due to radiation) is identical before and after the changes. Ida reports a very small effect on runaway generation in her more complex self-consistent iTER-like simulations. Since previously-unstable simulations became stable after the fix, I suspect that there was a subtle contribution to W_binding that we forgot to compensate for in the old model.


hoppe93 (approved):
I think it looks great!

PR #75, opened 2020-10-24
Useful formulas

hoppe93:
I added the module ``DREAM.Formulas`` where we can gather various useful formulas. At the moment, the module contains things such as thermal collision times, E_c, E_D, the Connor-Hastie runaway rate etc.


py/DREAM/Formulas/PlasmaParameters.py
```diff
@@ -0,0 +1,116 @@
...
+
+def getTauEETh(T, n):
+    """
+    Calculates the thermal electron-electron collision frequency.
+
+    :param float T: Plasma temperature (eV).
+    :param float n: Plasma density (m^-3).
+    """
+    mc2  = scipy.constants.physical_constants['electron mass energy equivalent in MeV'][0] * 1e6
+    vth2 = 2*T/mc2
```
Embreus, line 67: could do vth = getBetaThermal(T) here instead to ensure consistent defintions

Embreus (approved):
Love it! Nice to have such a class to collect useful formulas.

py/DREAM/Formulas/PlasmaParameters.py
hoppe93: Good point! I just added ``getBetaThermal`` today and only replaced it in the tests, but forgot about ``PlasmaParameters.py`` :P Updated now though!

PR #76, opened 2020-10-26
Rechester-Rosenbluth heat transport

hoppe93:
Implementation of Rechester-Rosenbluth heat transport


Embreus:
I have looked through all changes, and have no further suggestions for modifications before merging!

PR #77, opened 2020-10-29
Screening

hesslow:
Implement the iterative Eceff calculation from the PPCF paper. Divide Runaway-fluid class into three classes. Update tests in RunawayFluid.cpp.


include/DREAM/Equations/EffectiveCriticalField.hpp
```diff
@@ -0,0 +1,81 @@
...
+namespace DREAM {
+    class EffectiveCriticalField {
+    public:
+        struct ParametersForEceff {FVM::RadialGrid *rGrid; SlowingDownFrequency *nuS; PitchScatterFrequency *nuD; FVM::fluxGridType fgType; 
+                            gsl_integration_workspace *gsl_ad_w;
+                            gsl_min_fminimizer *fmin; CollisionQuantity::collqty_settings *collSettingsForEc;
+                            CollisionQuantity::collqty_settings *collQtySettings; gsl_root_fsolver *fsolve; 
+                            OptionConstants::collqty_Eceff_mode Eceff_mode;
+                            IonHandler *ions;
+                            CoulombLogarithm *lnLambda;};
```
hoppe93, line 26: I think this struct could be declared a bit more nicely, with one type per line (and the ``{`` ending its line). You can also declare ``collSettingsForEc`` and ``collQtySettings`` on the same line using
```
CollisionQuantity::collqty_settings *collSettingsForEc, *collQtySettings;
```

src/Equations/AnalyticDistribution.cpp
```diff
@@ -0,0 +1,120 @@
...
+using namespace DREAM;
+
+/**
+ * Constructor.
+ *
+ */
+AnalyticDistribution::AnalyticDistribution(FVM::RadialGrid *rGrid, PitchScatterFrequency *nuD, OptionConstants::collqty_Eceff_mode Eceff_mode) 
+: rGrid(rGrid), nuD(nuD), Eceff_mode(Eceff_mode){}
+
+// ok, maybe re-name the constant so it doen't belong to Eceff?
```
hoppe93, line 20: You could add a proper comment for the method here. Here's a template:
```
/**
 * DESCRIPTION
 *
 * ir:         DESC
 * xi0:        DESC
 * p:          DESC
 * Eterm:      DESC
 * inSettings: DESC
 * gsl_ad_w:   DESC
 */
```

src/Equations/AnalyticDistribution.cpp
```diff
@@ -0,0 +1,120 @@
...
+real_t AnalyticDistribution::evaluatePitchDistribution(len_t ir, real_t xi0, real_t p, 
+real_t Eterm, CollisionQuantity::collqty_settings *inSettings, gsl_integration_workspace *gsl_ad_w){
```
hoppe93, line 22: I find it a bit confusing to split the argument list in this way (with same indentation for arguments and method head). I would rather write it as
```
real_t AnalyticDistribution::evaluatePitchDistribution(
    len_t ir, real_t xi0, real_t p, real_t Eterm,
    CollisionQuantity::collqty_settings *inSettings, gsl_integration_workspace *gsl_ad_w
) {
```

src/Equations/AnalyticDistribution.cpp
```diff
@@ -0,0 +1,120 @@
...
+//    printf("Eceffmode = %d\n",Eceff_mode);
```
hoppe93, line 23: Remove?

src/Equations/AnalyticDistribution.cpp
```diff
@@ -0,0 +1,120 @@
...
+}
+
+/**
+ * Calculates the (semi-)analytic pitch-angle distribution predicted in the 
+ * near-threshold regime, where the momentum flux is small compared 
+ * to the characteristic pitch flux, and we obtain the approximate 
+ * kinetic equation phi_xi = 0.
+ */
+real_t AnalyticDistribution::evaluateAnalyticPitchDistribution(len_t ir, real_t xi0, real_t p, real_t Eterm, 
+CollisionQuantity::collqty_settings *inSettings, gsl_integration_workspace *gsl_ad_w){
```
hoppe93, line 52: (Same comment as for ``evaluatePitchDistribution()`` above)

src/Equations/RunawayFluid.cpp
```diff
@@ -229,7 +233,10 @@ void RunawayFluid::GridRebuilt(){
     nuD->GridRebuilt();    
 }
 
+<<<<<<< HEAD
+=======
 
+>>>>>>> master
```
hoppe93, line 239: These are remains from a merge conflict and shouldn't compile?

src/Equations/RunawayFluid.cpp
```diff
@@ -408,6 +417,7 @@ void RunawayFluid::FindPExInterval(real_t *p_ex_guess, real_t *p_ex_lower, real_
 
 
 
+>>>>>>> master
```
hoppe93, line 420: Remove

src/IonHandler.cpp
```diff
@@ -230,7 +230,7 @@ real_t IonHandler::evaluateFreeElectronDensityFromQuasiNeutrality(len_t ir){
 /**
  * Returns the bound electron density n_bound.
  *
- * nbound: If NOT 'nullptr', this array contains the free electron
+ * nbound: If NOT 'nullptr', this array contains the free electron // I think "bound" should be "free" here, right?
```
hoppe93, line 233: Could we perhaps clarify this comment-in-comment before merging?

tests/cxx/tests/DREAM/MeanExcitationEnergy.cpp
```diff
@@ -0,0 +1,181 @@
...
+    2.68494e-04, 
+    3.23288e-04, 
+    6.90021e-04, 
+    3.69277e-04, 
+    4.29551e-04, 
+    1.55969e-03, 
+    7.04502e-04, 
+    8.06097e-04, 
+    3.45034e-03,
+    1.01486e-02
```
hoppe93, line 160: I would put all of these on one line and place the comment on the previous just above it, and align the element names with the values.

src/Equations/RunawayFluid.cpp
```diff
@@ -966,6 +964,7 @@ real_t RunawayFluid::testEvalU(len_t ir, real_t p, real_t Eterm, CollisionQuanti
 }
 
 /**
+>>>>>>> master
```
hoppe93, line 967: Remove

src/Equations/EffectiveCriticalField.cpp
```diff
@@ -0,0 +1,399 @@
...
+    case OptionConstants::COLLQTY_ECEFF_MODE_EC_TOT : { // or COLLQTY_ECEFF_MODE_NOSCREENING to be consistent with 
+    // for example GetConnorHastieField_NOSCREENING?
+        if(collQtySettings->collfreq_type==OptionConstants::COLLQTY_COLLISION_FREQUENCY_TYPE_COMPLETELY_SCREENED){
+            for(len_t ir=0; ir<nr; ir++)
+                effectiveCriticalField[ir] = Ec_free[ir];
+        }else{
+            for(len_t ir=0; ir<nr; ir++)
+                effectiveCriticalField[ir] = Ec_tot[ir]; 
+        }
+        return;
```
hoppe93, line 73: Remove

src/Equations/EffectiveCriticalField.cpp
```diff
@@ -0,0 +1,399 @@
...
+    }
```
hoppe93, line 74: Add a ``break`` after the ``}``

src/Equations/EffectiveCriticalField.cpp
```diff
@@ -0,0 +1,399 @@
...
+            for(len_t ir=0; ir<nr; ir++)
+                effectiveCriticalField[ir] = Ec_free[ir];  
+        }else{
+            // calculate Eceff according to paper
+            for(len_t ir=0; ir<nr; ir++){
+                effectiveCriticalField[ir] = CalculateEceffPPCFPaper(ir);
+            }
+        }
+        return;
+    }
```
hoppe93, line 88: Remove and add ``break;`` after ``}``

src/Equations/EffectiveCriticalField.cpp
```diff
@@ -0,0 +1,399 @@
...
+    case OptionConstants::COLLQTY_ECEFF_MODE_SIMPLE : {
```
hoppe93, line 89: Move the ``{`` to the next line

src/Equations/EffectiveCriticalField.cpp
```diff
@@ -0,0 +1,399 @@
...
+            /**
+             * Initial guess: Eceff is between 0.9*Ec_tot and 1.5*Ec_tot
+             */
+            ELo = .9*Ec_tot[ir];
+            EUp = 1.5*Ec_tot[ir];
+            RunawayFluid::FindInterval(&ELo, &EUp, UExtremumFunc);
+            RunawayFluid::FindRoot(ELo,EUp, &effectiveCriticalField[ir], UExtremumFunc,fsolve);
+        }
+        return;
+    }
```
hoppe93, line 112: Remove and replace with ``break;`` after the ``}``

src/Equations/EffectiveCriticalField.cpp
```diff
@@ -0,0 +1,399 @@
...
+    default :
+        return;
```
hoppe93, line 114: Replace with ``break``

hoppe93 (changes requested):
Looks mostly really good, but there are a few things that absolutely must be changed before merging (and a few things which would be nice if you could fix :)

src/Equations/AnalyticDistribution.cpp
hesslow: @Embreus wrote this function, so I would suggest he takes over this fix.

src/Equations/AnalyticDistribution.cpp
hesslow: @Embreus wrote this function, so I would suggest he fixes it

src/Equations/RunawayFluid.cpp
hesslow: probably forgot to save the last change...

tests/cxx/tests/DREAM/MeanExcitationEnergy.cpp
hesslow: Do you mean like
const real_t TABULATED_MEAN_EXCITATION_ENERGIES[] = {
        //Ne0,            Ne1,         Ne5, ...       
        X,                  Y              Z, ...
    };

?

hoppe93: Essentially. Exactly like this:
```
// Ne0,      Ne1          Ne5
2.68494e-04, 3.23288e-04, 6.90021e-04, ...
```

Embreus:
Should the RunawayFluid.UFunc.cpp file be removed?

hoppe93 (approved):
Excellent!

include/DREAM/Equations/EffectiveCriticalField.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+        void CalculateEffectiveCriticalField(const real_t *Ec_tot, const real_t *Ec_free, real_t *effectiveCriticalField);
+        real_t CalculateEceffPPCFPaper(len_t ir);
+
+        static real_t FindUExtremumAtE(real_t Eterm, void *par);
+        static void FindPExInterval(real_t *p_ex_guess, real_t *p_ex_lower, real_t *p_ex_upper, real_t p_upper_threshold, 
+        UContributionParams *params);
+        static real_t UAtPFunc(real_t p, void *par);
+
+        //@@Linnea should be sent over to the runaway-fluid class for backwards compability.
```
Embreus, line 81: what stopping us?

src/Equations/EffectiveCriticalField.cpp
```diff
@@ -0,0 +1,396 @@
...
+            }else{
+                // calculate Eceff according to paper
+                for(len_t ir=0; ir<nr; ir++){
+                    effectiveCriticalField[ir] = CalculateEceffPPCFPaper(ir);
+                }
+            }
+        }
+        break;
+        case OptionConstants::COLLQTY_ECEFF_MODE_SIMPLE : 
+        {
```
Embreus, line 86: actually, how does this work? @hesslow @hoppe93 
does this cause SIMPLE to fallthrough to FULL (which is the desired behaviour)?

hoppe93: No, I would expect this to be invalid code. The case will fallthrough as long as there is no ``break`` between them. In this mode, it looks as if the ``FULL`` case is within the ``SIMPLE`` case, and thus not accessible to the switch statement? (Probably it is though, because otherwise I think the compiler should've complained)

Embreus: I changed this to something that is hopefully clearer now

hesslow: Hm, seems like I missed the C++17 updates for switch statements (if I understood correctly)? Thanks for fixing!

hesslow:
> Should the RunawayFluid.UFunc.cpp file be removed?

Yes, the functions are now in either `EffectiveCriticalField` or `AnalyticDistribution` (Google told me including c++ files like this was rather unconventional).

include/DREAM/Equations/EffectiveCriticalField.hpp
hesslow: Because I got the impression that these functions were not used anywhere else, so it was unnecessary. And I forgot to ask you. So now I ask!

Embreus:
I removed it after making this comment.

include/DREAM/Equations/EffectiveCriticalField.hpp
Embreus: I've removed it. :)

PR #80, opened 2020-11-01
Ripple

Embreus:
Implements resonant pitch-angle diffusion with the toroidal magnetic field ripple

### DIFF
--- a/py/DREAM/Settings/Equations/DistributionFunction.py
+++ b/py/DREAM/Settings/Equations/DistributionFunction.py
@@ -355,15 +355,16 @@ class DistributionFunction(UnknownQuantity):
             bc = self.boundarycondition
             if (bc != BC_F_0) and (bc != BC_PHI_CONST) and (bc != BC_DPHI_CONST):
                 raise EquationException("{}: Invalid external boundary condition set: {}.".format(self.name, bc))
-            ad_int_r = self.adv_interp_r
-            if (ad_int_r != AD_INTERP_CENTRED) and (ad_int_r != AD_INTERP_DOWNWIND) and (ad_int_r != AD_INTERP_UPWIND) and (ad_int_r != AD_INTERP_UPWIND_2ND_ORDER) and (ad_int_r != AD_INTERP_QUICK) and (ad_int_r != AD_INTERP_SMART) and (ad_int_r != AD_INTERP_MUSCL)  and (ad_int_r != AD_INTERP_OSPRE) and (ad_int_r != AD_INTERP_TCDF): 
-                raise EquationException("{}: Invalid radial interpolation coefficient set: {}.".format(self.name, ad_int_r))
-            ad_int_p1 = self.adv_interp_p1
-            if (ad_int_p1 != AD_INTERP_CENTRED) and (ad_int_p1 != AD_INTERP_DOWNWIND) and (ad_int_p1 != AD_INTERP_UPWIND) and (ad_int_p1 != AD_INTERP_UPWIND_2ND_ORDER) and (ad_int_p1 != AD_INTERP_QUICK) and (ad_int_p1 != AD_INTERP_SMART) and (ad_int_p1 != AD_INTERP_MUSCL)  and (ad_int_p1 != AD_INTERP_OSPRE) and (ad_int_p1 != AD_INTERP_TCDF):
-                raise EquationException("{}: Invalid p1 interpolation coefficient set: {}.".format(self.name, ad_int_p1))
-            ad_int_p2 = self.adv_interp_p2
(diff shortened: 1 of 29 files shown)


py/DREAM/Settings/Equations/DistributionFunction.py
```diff
@@ -190,6 +199,49 @@ def setInitialValue(self, f, r, p=None, xi=None, ppar=None, pperp=None):
...
+
+        if type(t) == list: t = np.array(t)
+        elif np.isscalar(t): t = np.array([float(t)])
+
+        if type(dB_B) == list:
+            dB_B = np.array(dB_B)
+        if type(dB_B) == float or dB_B.ndim == 1:
+            dB_B = np.ones((m.size, t.size, r.size)) * dB_B
+
+        if m.size != n.size:
```
Embreus, line 231: it would be nice to allow dB_B to be given as a list of length m.size, and in this method rescale it to the correct format as constant in t and r.

py/DREAM/Settings/Equations/DistributionFunction.py
```diff
@@ -67,6 +67,15 @@ def __init__(self, settings, name, grid,
         self.T0  = rT0
         self.rT0 = T0
 
+        # Ripple parameters
```
Embreus, line 70: would be nice to have the settings in RadialGrid instead, so that you don't have to specify these settings separately for f_hot and f_re

src/ConvergenceChecker.cpp
```diff
@@ -82,8 +83,11 @@ void ConvergenceChecker::DefineAbsoluteTolerances() {
  * name: Name of unknown quantity to get default absolute tolerance for.
  */
 real_t ConvergenceChecker::GetDefaultAbsTol(const string &name) {
+    real_t abstol_nre = 1e-10;
     if (name == OptionConstants::UQTY_N_RE)
-        return 1e12;
+        return abstol_nre;
+    else if ( name == OptionConstants::UQTY_J_RE)
+        return Constants::ec*Constants::c*abstol_nre;
```
Embreus, line 90: I made the j_re tolerance consistent with the n_re tolerance (which I also lowered a bit since I felt the default was a bit too coarse)

src/Equations/Kinetic/RipplePitchScattering.cpp
```diff
@@ -0,0 +1,175 @@
...
+        throw DREAMException(
+            "PitchScattermTerm: Please use the parameter 'deltaCoils' instead of "
+            "'nCoils' in cylindrical mode."
+        );
+}
+
+RipplePitchScattering::RipplePitchScattering(
+    FVM::Grid *grid, enum OptionConstants::momentumgrid_type mgtype,
+    const real_t deltaCoils, const len_t nModes, const int_t *m,
+    const int_t *n, DREAM::IonInterpolator1D *dB_B
```
Embreus, line 49: there was the question of whether we should rename IonInterpolator to MultInterpolator or similar?

src/OtherQuantityHandler.cpp
```diff
@@ -273,6 +276,20 @@ void OtherQuantityHandler::DefineQuantities() {
             real_t *ncold = this->unknowns->GetUnknownData(this->id_ncold);
             this->tracked_terms->T_cold_radiation->SetVectorElements(qd->StoreEmpty(), ncold);
         );
+    // Magnetic ripple resonant momentum
+    if (tracked_terms->f_hot_ripple_Dxx != nullptr) {
+        len_t nModes = tracked_terms->f_hot_ripple_Dxx->GetNumberOfModes();
+        DEF_FL("fluid/f_hot_ripple_m", "Magnetic ripple poloidal mode number for f_hot", qd->Store(this->tracked_terms->f_hot_ripple_Dxx->GetPoloidalModeNumbers()););
```
Embreus, line 282: if we move the toroidal field quantities to radialgrid, we could rename these to just fluid/ripple_m etc.

You would probably still want to check whether f_hot_ripple_Dxx _or_ f_re_ripple_Dxx are not nullptr, since in principle we may want to support running f_re without f_hot.

src/Settings/Equations/distribution.cpp
```diff
@@ -54,6 +54,23 @@ void SimulationGenerator::DefineOptions_f_general(Settings *s, const string& mod
 
     // Kinetic transport model
     DefineOptions_Transport(mod, s, true);
+
+    // Magnetic ripple effects
+    DefineOptions_f_ripple(mod, s);
```
Embreus, line 59: _RadialGrid_ripple

Embreus:
I think that this is looking pretty close to good-to-go now! I left a few wish-list comments, but nothing critical.

Issue #87, opened 2020-11-12
Physics-based preconditioner


hoppe93:
In particular MUMPS seems to sometimes have problems converging in certain scenarios. It is unclear exactly what the cause is, but a probable cause might be that the lack of normalizations in DREAM make the matrix ill-conditioned. One potential way of relieving this could be to simply rescale the matrix in order to make all matrix elements as close to unity as possible. This should be relatively straightforward to achieve via a transformation A' = P*A*P^-1 before inverting the matrix A, with P a (diagonal) rescaling matrix.

Embreus:
I wonder if you want to rescale rows and columns differently? Because the equation corresponding to a given unknown ID does not necessarily have the same scale as the unknown. For example, the equation for T_cold is in SI, so you would want to rescale it differently than how you would rescale T_cold.

Then, you could cast the equation

Ax = b

to

P A Q^-1 Qx = Pb

where the diagonal P contains scale factors for the equations, and Q for the unknowns. 

I guess that we want to have constant default values for each unknown id, and not rescale each radius separately depending on parameters? Then, we could for example rescale unknowns by:

- Densities (electrons, ions) and distribution functions: 1e20 (maybe n_re and f_re by something smaller, say 1e10?)
- Current densities: 1e6 (1 MA/m^2)
- Electric field, temperature, poloidal flux: 1

For the equations, we could scale by:

- Temperature: 1e6 (1 MJ/m^3)
- Ions: 1e26 (dn_i/dt with characteristic ionization times of 1e-6 seconds)
- otherwise same as unknowns?

Of course, if we don't care about scale factors <1e10 we could just rescale densities and distributions by 1e20 and call it a day.

Embreus:
I tried this with a DREAM jacobian matrix in matlab. By tweaking the scalefactors a bit I could get the matrix to become well-conditioned, and the time to solve Ax=b for a test vector b went down to 40% of the previous time, using essentially the scalefactors given in the previous comment.

Embreus:
An alternative way of choosing scalefactors could be to take them from the norm (over all elements of the same ID) of the unknowns in the previous iteration and vector of the present, respectively. That may be more robust, and in a way also simpler.

Embreus:
[...]

PR #104, opened 2020-11-30
Preconditioner

hoppe93:
This pull request implements a new preconditioner intended to improve the condition of the equation system matrices and solving #87. The preconditioner appears to be working as expected, although it does not appear to solve the issue with MUMPS where the equation system suddenly crashes for no apparent reason.


Embreus:
I've been through it in some detail and think it looks ready to go now!

Issue #31, opened 2020-08-24
Implement Linnea's Eceff formula


Embreus:
Currently in DREAM, there are three modes of evaluating Eceff
- Eceff = Ectot. Super fast, but not super accurate (up to a factor of ~2 error)
- Eceff approx. Really slow. Evaluates the Lehtinen theory prediction using a simplified pitch distribution. (agrees with accurate for cylindrical) 
- Eceff accurate. Ultra slow. Full Lehtinen theory prediction.

We should implement Linnea's efficient intermediate formula which does not account for trapping effects, but has radiation and screening effects. This would likely not be so slow that it bottlenecks the simulation, and is a lot more accurate than setting just Ectot.

Embreus:
I can also see a possible optimization of the full calculation which may have a huge effect.

Currently, we perform a set of fundamental bounce averages, such as {xi}, <xi0/xi>, {B^3} etc, for all xi0 at each momentum and E for which the function is sampled. Thus, we perform the same bounce average calculations ~a million times for each evaluation of Eceff. 

If we instead  precompute and tabulate these bounce averages on a reference xi0 grid, we can instead integrate the splines of the tabulated values. 
Indeed, since in the end we only need the integrals
int( {...} * exp(- A*(1-xi0))  dxi0 ) for three different bounce averages {...}, we may even opt to tabulate these integrals for a range of A values instead. Then, splining these for a range of tabulated values of A = 2e*Eterm/(p*nuD), the resulting "full" Eceff calculation would likely become lightning fast.

Embreus:
Fixed in #77

Embreus:
Reopened for the rework suggested above.

We could overload AnalyticDistribution::evaluateAnalyticPitchDistribution with a method taking only (real_t A, real_t xi0) as arguments which carries out the last couple of lines in the existing method.

In EffectiveCriticalField::UAtPFunc, there are three places where 3 gsl integrals are called. These three terms, which depend only on radius and A (it looks like it depends on p, but actually the p is just carried through and is never used, and so it can be set to p = 0), could be splined in nr different 1D splines each as function of A. 

[...]

PR #105, opened 2020-12-01
Screening

hesslow:
implemented part two of issue #31.


src/Equations/AnalyticDistributionRE.cpp
```diff
@@ -47,16 +73,13 @@ real_t distExponentIntegral(real_t xi0, void *par){
...
+){
     const real_t Bmin = rGrid->GetBmin(ir);
     const real_t Bmax = rGrid->GetBmax(ir);
-    const real_t B2avgOverBmin2 = rGrid->GetFSA_B2(ir);
-    real_t xiT = sqrt(1-Bmin/Bmax);
-    real_t E = Constants::ec * Eterm / (Constants::me * Constants::c) * sqrt(B2avgOverBmin2); 
-
-    real_t pNuD = p*nuD->evaluateAtP(ir,p,inSettings);    
-    real_t A = 2*E/pNuD;
+    real_t xiT = sqrt(1-Bmin/Bmax);
```
Embreus, line 82: instead use: 
real_t xiT = rGrid->GetXi0TrappedBoundary(ir)

src/Equations/AnalyticDistributionRE.cpp
```diff
@@ -92,15 +115,10 @@ CollisionQuantity::collqty_settings *inSettings, gsl_integration_workspace *gsl_
  * xi0/<xi> = 1 for passing and 0 for trapped (thus avoiding the 
  * need for the numerical integration).
  */
-real_t AnalyticDistributionRE::evaluateApproximatePitchDistribution(len_t ir, real_t xi0, real_t p, real_t Eterm, CollisionQuantity::collqty_settings *inSettings){
+real_t AnalyticDistributionRE::evaluateApproximatePitchDistributionFromA(len_t ir, real_t xi0, real_t A){
     const real_t Bmin = rGrid->GetBmin(ir);
     const real_t Bmax = rGrid->GetBmax(ir);
-    const real_t B2avgOverBmin2 = rGrid->GetFSA_B2(ir);
     real_t xiT = sqrt(1-Bmin/Bmax);
```
Embreus, line 121: rGrid->GetXi0TrappedBoundary(ir)

src/Equations/EffectiveCriticalField.cpp
```diff
@@ -57,6 +80,52 @@ bool EffectiveCriticalField::GridRebuilt(){
...
+    if ((Eceff_mode == OptionConstants::COLLQTY_ECEFF_MODE_SIMPLE) || (Eceff_mode == OptionConstants::COLLQTY_ECEFF_MODE_FULL)){
+        
+        this->EOverUnityContrib = new real_t*[nr];
+        this->SynchOverUnityContrib = new real_t*[nr];
+
+        this->gsl_parameters.EContribSpline = new gsl_spline*[nr]; 
+        this->gsl_parameters.SynchContribSpline = new gsl_spline*[nr];
+
+        this->gsl_parameters.EContribAcc = gsl_interp_accel_alloc(); // the accelerators cache values from the splines
+        this->gsl_parameters.SynchContribAcc = gsl_interp_accel_alloc();
```
Embreus, line 102: I don't quite remember (and I've previously stated otherwise), but I think these accelerator objects should only be assigned to one spline each.. so for example the EContribAcc should be a pointer to an array of nr accelerators

src/Equations/EffectiveCriticalField.cpp
```diff
@@ -327,67 +386,143 @@ real_t EffectiveCriticalField::UAtPFunc(real_t p, void *par){
         Bmin = rGrid->GetBmin(ir);
         Bmax = rGrid->GetBmax(ir);
     }
-    const real_t sqrtB2avgOverBavg = sqrt(rGrid->GetFSA_B2(ir)) / rGrid->GetFSA_B(ir);
     real_t xiT = sqrt(1-Bmin/Bmax);
```
Embreus, line 389: if(fluxGridType == ..._RADIAL
xiT = rGrid->GetXi0TrappedBoundary_fr(ir)
else
xiT = rGrid->GetXi0TrappedBoundary(ir)

Embreus (changes requested):
Left a few minor comments!

src/Equations/EffectiveCriticalField.cpp
hesslow: actually, I first thought you were wrong and implemented arrays of accelerators, but then I realized they could be reset so I thought that was what you meant all the time;) Isn't resetting in every radial point better? If I need to re-calculate Eceff I need to reset them anyway, so no need to save for later iterations.

Embreus: I thought you only needed to reset if the data behind the splines has changed?

hoppe93: The accelerators function as "memory" for looking up points on the x grid (in this case r?). When you reset the accelerator you wipe the memory which has been built during the past evaluations of the spline. Accelerators are useful because, aside from just multiplying/adding the spline coefficients, the other expensive part of the evaluation is to figure which set of coefficients (corresponding to one/several x points) to use. Since we can generally make very few assumptions about the x vector (I think GSL only assumes it is monotonically increasing/decreasing), these search algorithms tend to be very general and slow, which is why accelerators can make a big difference when evaluating splines.

hesslow: @Embreus yes, the data changes because I the same accelerator for all radii. So instead of a new accelerator I just reset it between different radial points. 
@hoppe93: we interpolate in A here

Embreus: I think it would be a good idea then to have arrays of accelerators, to avoid having to reset them.

PR #107, opened 2020-12-06
Initialize f_re consistently from n_re

hoppe93:
This branch adds support for initializing ``f_re`` from ``n_re``. If the hot-tail grid is disabled, ``f_re`` is automatically initialized from ``n_re``. By default, all particles are placed in ``p=pMin`` and ``xi=+/-1``, with the sign of ``xi`` being the same as the sign of the electric field ``E`` (or +1 if ``E=0``), but it is also possible to explicitly put particles in the desired ``xi`` direction, or place isotropically in ``p=pMin``.

This branch also contains a minor bug fix in softlib. Now, the ``SFile`` object flushes the HDF5 file handle right before closing, ensuring that the file is actually written immediately. Previously it was found that an unidentified bug sometimes causes DREAM to crash when ``dream_finalize()`` is called. The bug occurs when deinitializing MPI (in PETSc) and would sometimes lead to the HDF5 output file being corrupted. With this fix the HDF5 file appears to always be correctly written *before* DREAM exits.


Embreus (changes requested):
Looks pretty much good to go, as far as I can tell! It's a nice solution =) 

I think that we should just fix the trapping case before merging. Essentially a check before setting the initial values 
if( grid->IsNegativePitchTrappedIgnorableCell(ir, xiindex) ) 
    [ set mirrored point(s) to the same value ] 

for isotropic, this check is not necessary (since everything has the same value anyway)

Embreus:
I was looking to fix my comment, but realize that it's not really needed with the present implementation.. for xi=+/-1 they wouldn't be trapped for any sensible grid, and for 'isotropic' it's already correct.  It'll be more like something to keep in mind in the future, if we choose to extend the initialization.

PR #109, opened 2020-12-08
Refactored handling of advection interpolation

hoppe93:
In this commit I propose a unification of the advection interpolation interface which should make it easier to apply more generally to equations. In particular, it arose out of the perceived need to include advection interpolation in the transport equation for ``n_re`` and may be useful in other transport equations in the future.

(note that this is **NOT** a pull request to master, but rather just to the WIP branch ``SvenssonTransport``)


Embreus (approved):
I'm a huge fan!

PR #112, opened 2020-12-16
Bounceaverage

Embreus:
This branch contains a few fixes so that the advection, diffusion and 2kinetic tests succeed when using general analyticB grids.

In the end, the only proper error was that we gave a non-zero Vp to the cell containing the negative trapped-passing boundary (due to the overlap of the cell with the negative-pich _passing_ region). This was inconsistent, since all fluxes to this cell were moved to the mirrored cell and the distribution in the cell was not actually evolved via fluxes. Now Vp in this cell is set to zero instead, which ensures conservativity.

The branch also updates the conservativity unit test of DREAMTESTS::EquationTerm, which contained a few errors such that it previously only tested conservativity for a uniform distribution and with constant grid spacing. Now it should be general.


src/Settings/RunawayFluid.cpp
```diff
@@ -50,7 +50,7 @@ RunawayFluid *SimulationGenerator::ConstructRunawayFluid(FVM::Grid *g,
         Eceff_mode, ava_mode, compton_mode, compton_photon_flux
     );
 
-    delete cq;
+//    delete cq;
```
hoppe93, line 53: Is this correct? If it is, maybe just remove?

tests/cxx/tests/FVM/GeneralDiffusionTerm.cpp
```diff
@@ -25,23 +25,27 @@ GeneralDiffusionTerm::GeneralDiffusionTerm(DREAM::FVM::Grid *g, const real_t v)
...
-    #define SETD21(V) if (i < np1) D21(ir,i,j) = (V)
-    #define SETD22(V) if (i < np1) D22(ir,i,j) = (V)
-
-    for (len_t ir = 0; ir < nr; ir++) {
-        auto *mg = this->grid->GetMomentumGrid(ir);
+    #define SETDRR(V) if (ir < nr && i < np1 && j < np2 && ir != 0) Drr(ir,i,j) = (V)
+    #define SETD11(V) if (ir < nr && i < np1 && j < np2 && i  != 0) D11(ir,i,j) = (V)
+    #define SETD12(V) if (ir < nr && i < np1 && j < np2 && i  != 0) D12(ir,i,j) = (V)
+    #define SETD21(V) if (ir < nr && i < np1 && j < np2 && j  != 0) D21(ir,i,j) = (V)
+    #define SETD22(V) if (ir < nr && i < np1 && j < np2 && j  != 0) D22(ir,i,j) = (V)
```
hoppe93, line 39: Why was it necessary to extend these if statements?

hoppe93 (commented):
Looks good! I just had a few very minor questions and comments.

src/Settings/RunawayFluid.cpp
Embreus: right, this is not completely finished. In the RunawayFluid constructor we create a new collqty settings objects that is to be used in all calculations. I want to create that here in the settings file instead, and generate nuS and nuD using that (previously we apparently got a valgrind error because cq (provided to nuS and nuD) was deleted, which was scary)

tests/cxx/tests/FVM/GeneralDiffusionTerm.cpp
Embreus: I changed it so that the flux in the terms is explicitly set to zero on the boundary, since it is strictly only then that the equation should be conservative. The alternative is to have non-zero fluxes on the boundary and then enforce a Dirichlet boundary condition where those values are ignored, but that required more finesse to work well I felt (for some reason I cannot immediately recall). If you prefer the other way, I can look into reworking it (back to the way it was originally designed).

PR #117, opened 2021-01-21
Fluid hottail runaway rate

Embreus:
Implements the hot-tail runaway rate, as well as infrastructure surrounding analytic distributions that will be used in future equation terms (upcoming: AnalyticalMomentQuantity).

This first implementation is missing some jacobian contributions to the HottailRateTerm, but most of it is there and it seems to work quite well.

Also contains fixes for the nnz_jac calculation in some equation terms.


fvm/Equation/AdvectionInterpolationCoefficient.cpp
```diff
@@ -486,14 +486,16 @@ std::function<real_t(int_t)> AdvectionInterpolationCoefficient::GetYFunc(len_t i
...
  */
 void AdvectionInterpolationCoefficient::SetNNZ(adv_interpolation adv_i){
     bool isFirstOrder = ( (adv_i==AD_INTERP_CENTRED) || (adv_i==AD_INTERP_DOWNWIND) 
                        || (adv_i==AD_INTERP_UPWIND) );
     if(isFirstOrder)
-        nnzPerRow = 6*1+1; // = 7
+        nnzPerRow = 2; // = 7
     else
-        nnzPerRow = 6*STENCIL_WIDTH+1; // = 13
+        nnzPerRow = 4; // = 13
```
hoppe93, line 498: What's the motivation for this change?

hoppe93: Or, this is just for a single coefficient(/coordinate), right? Then it makes sense :P

py/DREAM/Settings/Equations/DistributionFunction.py
```diff
@@ -338,6 +356,11 @@ def todict(self):
             data['transport'] = self.transport.todict()
             data['fullIonJacobian'] = self.fullIonJacobian
 
+        if self.mode != DISTRIBUTION_MODE_NUMERICAL:
```
hoppe93, line 359: Why not ``self.mode == DISTRIBUTION_MODE_ANALYTICAL``?

fvm/Equation/AdvectionInterpolationCoefficient.cpp
Embreus: Yep, now off-diagonal contributions per coordinate! Will clean up the comments.

py/DREAM/Settings/Equations/DistributionFunction.py
Embreus: The intent was to later have multiple analytic modes (with/without contributions to n_hot and j_hot for example)

src/Equations/AnalyticDistributionHottail.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+    return true;
+}
+
+/**
+ * Evaluate the pitch-angle averaged energy distribution
+ */
+real_t AnalyticDistributionHottail::evaluateEnergyDistribution(len_t ir, real_t p, real_t *dfdp, real_t *dfdr){
+    if(type == OptionConstants::UQTY_F_HOT_DIST_MODE_NONREL){
+        real_t tau = unknowns->GetUnknownData(id_tau)[ir];
```
hoppe93, line 63: What happens here if ``id_tau`` was not initialized on line 17?

src/Settings/EquationSystem.cpp
```diff
@@ -136,8 +136,11 @@ void SimulationGenerator::ConstructEquations(
         CollisionQuantityHandler *cqh = ConstructCollisionQuantityHandler(re_type, runawayGrid, unknowns, ionHandler, s);
         eqsys->SetRunawayCollisionHandler(cqh);
     }
-    RunawayFluid *REF = ConstructRunawayFluid(fluidGrid,unknowns,ionHandler,re_type,s);
+    AnalyticDistributionRE *distRE = nullptr;
+    AnalyticDistributionHottail *distHT = nullptr;
+    RunawayFluid *REF = ConstructRunawayFluid(fluidGrid,unknowns,ionHandler,distRE,distHT,re_type,s);
     eqsys->SetREFluid(REF);
+    eqsys->SetAnalyticDists(distRE, distHT);
```
hoppe93, line 143: I think that in some other places I have instead passed ``eqsys`` to the ``ConstructXXX()`` method so that it can call things like ``eqsys->SetAnalyticDist(...)`` instead of having these lines here, and passing in variables which are really just additional return parameters.

src/Settings/Equations/f_re.cpp
```diff
@@ -95,9 +95,9 @@ void SimulationGenerator::ConstructEquation_f_re(
 			runawayGrid, eqsys->GetHotTailGrid(), runawayGrid, eqn_f_hot,
 			id_f_hot, id_f_re, FVM::BC::PXiExternalKineticKinetic::TYPE_UPPER
 		));
-    } else {}
-        // The fluid runaway source terms are the "boundary condition" at p = pMin
-
+    } // else {
+      //    The fluid runaway source terms are the "boundary condition" at p = pMin
+      // }
```
hoppe93, line 100: Why this change? Didn't it look better (more aligned) before?

src/Settings/EquationSystem.cpp
Embreus: Ah, that sounds like a good idea. I much prefer that solution!

tests/cxx/tests/DREAM/RunawayFluid.cpp
```diff
@@ -321,7 +325,13 @@ bool RunawayFluid::CompareEceffWithTabulated(){
     real_t delta3 = abs(Eceff3-TabulatedEceff3)/TabulatedEceff3;
     real_t threshold = 1e-3; 
     bool success = (delta1 < threshold) && (delta2 < threshold) && (delta3 < threshold);
-
+    if(delta1>threshold)
+        std::cout << "delta1:" << delta1 << std::endl;
+    if(delta2>threshold)
+        std::cout << "delta2:" << delta2 << std::endl;
+    if(delta3>threshold)
+        std::cout << "delta3:" << delta3 << std::endl;
```
hoppe93, line 333: Should these remain, or are they just debug statements? If they should remain, maybe it would be better to use ``this->PrintInfo("delta1: %e", delta1)`` etc?

src/DiagonalPreconditioner.cpp
```diff
@@ -174,7 +176,7 @@ void DiagonalPreconditioner::SetDefaultScalings() {
         } else {
             DREAM::IO::PrintWarning(
                 "DiagonalPreconditioner: Unrecognized unknown '%s'. Unknown "
-                "will not be preconditioned."
+                "will not be preconditioned.", name
```
hoppe93, line 179: ``name`` -> ``name.c_str()``. This appears to be the reason the LLVM test fails.

hoppe93 (changes requested):
We just need to fix the LLVM error also I think. Then it should be pretty ready to go (although you might want to have a look at my other comments first)

PR #125, opened 2021-02-08
Jacobians for boundary conditions and memory optimizations

hoppe93:
This pull request contains a number of improvements which can be put in two categories:

**New jacobian contributions for boundary conditions**
The various advection and diffusion boundary conditions have until now only received contributions to the main diagonal blocks of the jacobian matrix due to some difficulties with building the off-diagonal blocks. In this pull request, the off-diagonal blocks have been added in a fashion similar to the regular ``AdvectionTerm`` and ``DiffusionTerm`` operators which can greatly improve convergence in many scenarios.

**Memory optimizations**
It has often been observed that the first and second Newton iterations take significantly more time that subsequent iterations. This has been due to, in part, too little memory being allocated for various operators. All such errors found until now have been fixed none of the tests currently run indicate any allocation errors.

Problems relating to memory usage with the external linear solvers have also been investigated and mostly fixed.


fvm/Equation/EquationTerm.cpp
```diff
@@ -60,3 +60,23 @@ bool EquationTerm::GridRebuilt() {
...
 }
 
+/**
+ * Returns true if derivId is in the derivIdsJacobian vector.
+ *
+ * derivId:    ID of unknown quantity to differentiate w.r.t.
+ * nMultiples: OUTPUT. If not 'nullptr', will contain the number
+ *             of multiples of the quantity with respect to which
+ *             the derivative was taken.
+ */
```
Embreus, line 70: ah, thanks for cleaning this up

fvm/Matrix.cpp
```diff
@@ -57,20 +57,24 @@ void Matrix::Construct(
     this->m = m;
     this->n = n;
 
-    MatCreate(PETSC_COMM_WORLD, &(this->petsc_mat));
+    /*MatCreate(PETSC_COMM_WORLD, &(this->petsc_mat));
```
Embreus, line 60: can some of this debug be cleaned up?

py/DREAM/Output/EquationSystem.py
```diff
@@ -27,6 +27,7 @@ class EquationSystem:
         'j_tot':        CurrentDensity,
         'n_cold':       FluidQuantity,
         'n_hot':        FluidQuantity,
+        'N_i':          FluidQuantity,
```
Embreus, line 30: think I've commented this already, but I'm not sure if this (and W_i) should be fluidquantities, but perhaps rather some form of new ion quantity?

Embreus (changes requested):
I guess we just need to look into the remaining mallocs now! looking fantastic otherwise

py/DREAM/Output/EquationSystem.py
Embreus: Actually, this change made Ida's script crash with the following interface backtrace:

> Traceback (most recent call last):
> 
>   File "./run_scan2.py", line 340, in <module>
>     runiface(ds1,'out_1init.h5',quiet=True)
>   File "../../py/DREAM/runiface.py", line 73, in runiface
>     obj = DREAMOutput(outfile)
>   File "../../py/DREAM/DREAMOutput.py", line 44, in __init__
>     self.load(filename=filename, path=path)
>   File "../../py/DREAM/DREAMOutput.py", line 87, in load
>     self.eqsys = EquationSystem(od['eqsys'], grid=self.grid, output=self)
>   File "../../py/DREAM/Output/EquationSystem.py", line 57, in __init__
>     self.setUnknowns(unknowns)
>   File "../../py/DREAM/Output/EquationSystem.py", line 121, in setUnknowns
>     self.setUnknown(name=uqn, data=unknowns[uqn], attr=attr)
>   File "../../py/DREAM/Output/EquationSystem.py", line 101, in setUnknown
>     o = self.SPECIAL_TREATMENT[name](name=name, data=data, attr=attr, grid=self.grid, output=self.output)
>   File "../../py/DREAM/Output/FluidQuantity.py", line 27, in __init__
>     raise Exception("Unrecognized shape of data for '{}': {}. Expected (nt, nr) = ({}, {}).".format(name, data.shape, grid.t.size, grid.r.size))
> Exception: Unrecognized shape of data for 'N_i': (6, 3, 20). Expected (nt, nr) = (6, 20).
>

PR #128, opened 2021-02-22
Feature/lazyload

hoppe93:
Add support for loading simulation output data lazily in the output interface. Instead of opening and reading all output data into memory, the ``DREAMOutput`` object can now simply open the HDF5 file and only read data when needed. This both reduces the time it takes to load the output file and can significantly reduce the memory footprint. Lazy loading of HDF5 files is now the default, and the old behaviour (where all data was loaded immediately) can be used by providing an additional parameter when constructing the ``DREAMOutput`` object.


Embreus (approved):
I’ve got no suggestions for improvements, looks great!

PR #130, opened 2021-03-02
Heat diffusion

hoppe93:
Implementation of a new heat diffusion operator which contains a factor ``(3/2)*n_cold``, thus transport ``W_cold`` in the same way as it is done in GO.


Embreus:
for what it's worth, this looks really good!

PR #131, opened 2021-03-29
Backup inverter

hoppe93:
This branch implements a possibility to switch from a "main" linear system solver to a "backup" linear system solver when the main solver fails for some reason. This has been implemented in response to the more advanced solvers sometimes failing to converge in certain scenarios. With these changes, DREAM can then switch to the very robust (but slow) PETSc LU solver, while using faster solvers whenever possible.

In addition to the backup solver, the signature of the ``SetJacobianBlock()`` and associated methods have been updated to return ``bool``, indicating whether method has set any elements (corresponding to checking if the term depends on the specified unknown with ID ``derivId``).


Embreus (approved):
Looks fabulous!

PR #132, opened 2021-04-16
Feature/numericmag

hoppe93:
Support for arbitrary (2D) numeric magnetic fields in DREAM. Benchmark to analytic toroidal magnetic field is available. The implementation has also been tested on actual numerical data from ASDEX Upgrade and TCV to verify that the plasma volume is calculated accurately.


fvm/Grid/FluxSurfaceAverager.cpp
```diff
@@ -617,15 +624,21 @@ void FluxSurfaceAverager::FindThetas(
...
         // otherwise the solution is mirrored
         } else {
             *theta1 = -*theta2;
             return;
         }
-    } else // if not symmetric, look for solution in the remaining interval
-        x_lower = theta_Bmax-2*M_PI;
+    } else { // if not symmetric, look for solution in the remaining interval ("lower half-plane")
+        x_upper = theta_Bmin;
+        x_lower = theta_Bmax - (theta_Bmin<theta_Bmax ? 2*M_PI : 0);
```
hoppe93, line 635: For theta_Bmin > theta_Bmax (i.e. Bmin in lower half-plane), this does not live up to the specification given in the header of the method (the resulting interval would be on [theta_Bmax, theta_Bmin]). However, if I change it to live up to the definition, the code crashes elsewhere. So what should this method return if theta_Bmax < theta_Bmin?

fvm/Grid/NumericBRadialGridGenerator.cpp
```diff
@@ -0,0 +1,505 @@
...
+ * Calculate |\nabla r|^2 at the given poloidal angle.
+ *
+ * r:     Minor radius.
+ * theta: Poloidal angle.
+ */
+real_t NumericBRadialGridGenerator::NablaR2AtTheta(
+    const real_t r, const real_t theta
+) {
+	if (r == 0)
+		return 1.0;
```
hoppe93, line 464: @Embreus This line only works in the very special case of kappa=1. Any idea about how nabla-r could be evaluated in r=0 in numerical geometry?

hoppe93:
I am thinking about also adding support for the netCDF files we've received from SPARC (which I believe are more or less raw output from EFIT). Will probably happen in the future, on a separate branch, though.

fvm/Grid/FluxSurfaceAverager.cpp
Embreus: is there an example that demonstrates this crash? I am not entirely sure of what the issue is.. but I can easily imagine that this caused problems. I remember that I was looking into this briefly when I played around with the numericmag test, but didn't have time to come up with a good solution.

fvm/Grid/NumericBRadialGridGenerator.cpp
Embreus: annoyingly, I don't. However, does anything depend on the value coefficients take in r=0? Since the jacobian is 0 in this point, I imagine that all contributions from this point naturally vanish. Not the most comforting answer... Perhaps extrapolate from two small non-zero radii?

fvm/Grid/FluxSurfaceAverager.cpp
hoppe93: Looking at this again, I'm not sure what the issue would be (other than that the comment is specific to the case theta_Bmin<theta_Bmax). If you want to test you should be able to by saving the settings for the "numerical" run of the "numericmag" physics test. Otherwise, since this seems to work, we could probably just leave it.

fvm/Grid/NumericBRadialGridGenerator.cpp
hoppe93: Good point! Extrapolation would probably be a bit tricky (not super, but...), and if the value is anyway never going to be used it feels like a waste of time. Should add a comment though...

PR #137, opened 2021-05-31
Svensson transport

andsunds:
This is the "production speced" version of SvenssonTransport, used in the SPARC paper.

Remaining wish list:
 - Clean up code
 - Documentation on the usage of the code, and perhaps also on the theory
 - (Do-over of the implementations of the analytical distributions to use the dedicated class for that instead of home-brewed versions)


hoppe93:
I think I have seen a TeX document describing the implementation while in your office a few times. Could that be correct? Is there any particular reason that we shouldn't include that document in the GitHub repository?

andsunds:
Yes, there is a TeX document I used for converting the theory in Svensson et al. (2020), and I can include it. Although I think that at the moment, the document is lacking in some of the unit conversion, so that equations in the document does not correspond to the implemented ones. I'll have a look when I have time.

hoppe93:
I talked to Ola a bit about the analytical distribution functions, and he thinks that integrating them with SvenssonTransport would require a significant amount of work since they are currently designed to work with analytical weight functions. He thinks it's better to use the current dedicated implementation in SvenssonTransport.

So if we can just clean up the code on this branch, get the theory notes and add the online documentation we should be good to go. I will start with the online docs and then move on to code cleaning. If you could get some theory notes up @andsunds that would be great!

include/DREAM/Equations/Fluid/SvenssonTransport.tcc
```diff
@@ -0,0 +1,429 @@
...
+            // The variable to be added to
+            real_t pIntCoeff = 0;
+            
+            this->EvaluateIntegrand(ir);
+            
+            // The trapz integration in p
+            for (len_t i = 0; i < this->np-1; i++) {
+                pIntCoeff += 0.5 * (this->integrand[i] + this->integrand[i+1])
+                    * (this->p[i+1] - this->p[i]);
+                // YYY Jacobian??? * this->grid->GetVp(ir,i,0);
```
hoppe93, line 167: What does this comment mean?

hoppe93 (approved):
Looks great!

PR #141, opened 2021-06-28
Spi

vaoskar:
This pull request contains functions calculating some geometrical quantities necessary for SPI simulations to work with analytic B-geometry

### DIFF
--- a/doc/notes/SPIcoordinates.tex
+++ b/doc/notes/SPIcoordinates.tex
@@ -1,7 +1,7 @@
 \documentclass{notes}
 
 \title{Geometrical considerations for the SPI module}
-\author{Mathias Hoppe}
+\author{Mathias Hoppe and Oskar Vallhagen}
 
 
 \begin{document}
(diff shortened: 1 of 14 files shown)


doc/notes/SPIcoordinates.tex
```diff
@@ -87,16 +87,49 @@
...
+		R = R_0 + \Delta (r) + r[\cos{\theta}\cos{(\delta(r)\sin{\theta})} - \sin{\theta}\sin{(\delta(r)\sin{\theta})}].
+		\label{eq:RAnalyticB}
+	\end{equation}
+	From the equation for $Z$, we have
+	\begin{align}
+		\sin{\theta} &= \frac{Z}{r\kappa(r)}\nonumber \\
+		\cos{\theta} &= \pm \sqrt{1-\left(\frac{Z}{r\kappa(r)}\right)^2}.
+		\label{eq:SinCosAnalyticB}
+	\end{align}
+	It remains now to determine the sign on $\cos{\theta}$. We do this by calculating the major radius $R_\mathrm{crit}$ where $\cos{\theta}=0$. As the major radius is a monotonically increasing function of $\cos{\theta}$ {\color{red}(which it always is, right?)}, if $R>R_\mathrm{crit}$, $\cos{\theta}>0$, and vice versa. The critical major radius is calculated by noting that when $\cos{\theta}=0$, we have $\sin{\theta}=1$, so that
```
hoppe93, line 123: Technically R can vary non-monotonically with cos(theta), but this would most likely violate other assumptions in our bounce-averaged formulation (we assume that B varies monotonically between the two extrema on a flux surface). For fun, I created the attached figure, which has delta=1.5 and a non-monotonic R.
ExcessiveDelta

examples/SPI/D_Ne_SPI_ITER.py
```diff
@@ -99,10 +100,39 @@
 B0 = 5.3            # magnetic field strength in Tesla
 
 # Set up radial grid
+R0 = 6.2
+kappa = 1.2
+
+ds.radialgrid.setType(RGrid.TYPE_ANALYTIC_TOROIDAL)
```
hoppe93, line 106: While it doesn't appear to be a problem for this particular setup, keep in mind that if solving a kinetic equation in toroidal geometry one should really use a appropriately tuned pitch grid (i.e. use ``MomentumGrid.setTrappedPassingBoundaryLayerGrid()`` in the Python settings interface)

fvm/Grid/AnalyticBRadialGridGenerator.cpp
```diff
@@ -358,36 +358,342 @@ void AnalyticBRadialGridGenerator::InterpolateInputProfileToGrid(
...
+ * the maximum and minimum r for which the profile is defined. This is useful for the 
+ * GetRThetaPhiFromCartesian() and GetGradRCartesian() functions used for e.g. the SPI
+ * module in DREAM, as these functions must be able to handle coordinates outside the plasma
  *
- * (The Cartesian coordinate system is oriented such that x and y span
- * the poloidal plane. The origin of x and y is the magnetic axis.)
  */
-void AnalyticBRadialGridGenerator::GetRThetaFromCartesian(real_t* /*r*/, real_t* /*theta*/,
-    real_t /*x*/, real_t /*y*/, real_t /*z*/, real_t /*lengthScale*/, real_t /*startingGuessR*/
+real_t AnalyticBRadialGridGenerator::InterpolateInputProfileSingleExtrap(real_t r,
```
hoppe93, line 367: Wouldn't it be nice to implement helper methods for interpolating specifically in the various shaping parameters which utilize this method? E.g. a method ``InterpolateInputElongation(real_t r)``, which just returns ``kappa`` in the requested point ``r`` (and same for the other shaping parameters).

fvm/Grid/AnalyticBRadialGridGenerator.cpp
```diff
@@ -358,36 +358,342 @@ void AnalyticBRadialGridGenerator::InterpolateInputProfileToGrid(
...
+    const len_t nProvided, const real_t *xProvided, const real_t *xProvided_r,
+    gsl_spline *spline_x, gsl_interp_accel *spline_acc
 ) {
-	throw FVMException("AnalyticBRadialGridGenerator: This module is currently incompatible with the SPI module.");
+    if(nProvided==1){
+        return xProvided[0];
+    } else {
+        if(r>xProvided_r[nProvided-1])
+            return gsl_spline_eval(spline_x, xProvided_r[nProvided-1], spline_acc);
```
hoppe93, line 375: I think the GSL spline is defined so that if evaluated in a point contained in ``xProvided_r``, it will return exactly the corresponding value in ``xProvided`` (i.e. the spline passes through all given points exactly).

fvm/Grid/AnalyticBRadialGridGenerator.cpp
```diff
@@ -358,36 +358,342 @@ void AnalyticBRadialGridGenerator::InterpolateInputProfileToGrid(
...
+	// Determine the quadrant
+	len_t quadrant;
+	if(RMinusR0>RMinusR0_crit && y>=0)
+	    quadrant=1;
+	else if(RMinusR0<RMinusR0_crit && y>=0)
+	    quadrant=2;
+    else if(RMinusR0<RMinusR0_crit && y<0)
+        quadrant=3;
+    else if(RMinusR0>RMinusR0_crit && y<0)
+        quadrant=4;
```
hoppe93, line 479: You should configure your text editor to only use either tabs OR spaces. The current mixture of tabs and spaces (even on the same line) is bound to lead to confusion at some point.

hoppe93: (in Python it's of course even worse, but I presume you have some way of ensuring consistency there?)

hoppe93 (approved):
Looks pretty good! It would be nice if you could add the convenience methods to AnalyticBRadialGridGenerator, but otherwise it should be good to go.

hoppe93 (approved):

PR #145, opened 2021-09-23
Iontransport

vaoskar:
It might not be fully fit for merging at the moment, but I open the pull request anyway so you can have a look at it


fvm/Equation/AdvectionInterpolationCoefficient.cpp
```diff
@@ -14,11 +14,13 @@ using namespace std;
  * Constructor.
  */
 AdvectionInterpolationCoefficient::AdvectionInterpolationCoefficient(Grid*g, fluxGridType fluxGridType, 
-        adv_bc bc_l, adv_bc bc_u){
+        adv_bc bc_l, adv_bc bc_u, len_t offset){
```
hoppe93, line 17: Could you add a comment above describing the various input parameters? Something like:

```
g:            Grid on which the advection term lives.
fluxGridType: Coordinate in which we will interpolate (radius, p1 or p2).
bc_l:         Boundary condition to assume on the lower boundary.
bc_u:         Boundary condition to assume on the upper boundary.
offset:       For quantities with multiples, the offset into the quantity for
              which interpolation should be done (e.g. ion charge state
              index).
```

fvm/Equation/DiffusionTerm.set.cpp
```diff
@@ -44,12 +44,15 @@
...
-
                 // Trapping BC: even if the cell is not ignorable, it may still 
                 // be such that the radial flux should be mirrored 
-                if(!isNegativeTrappedRadial && ( Drr(ir, i, j, drr) || Drr(ir+1, i, j, drr) ) ){
+                
+                // For the Ions in DREAM the transient term is defined such that the 
+                // diffusion coefficients should must be negative to get the correct
+                // sign on the diffusion term, which is why we use abs(Drr) here. 
+                // This should however probably be reworked in a better way...
+                if(!isNegativeTrappedRadial && ( abs(Drr(ir, i, j, drr)) || abs(Drr(ir+1, i, j, drr)) ) ){
```
hoppe93, line 54: May want to explicitly give ``std::abs()``. The compiler may otherwise silently replace the ``std::abs()`` function with the old ``abs()`` function from the C standard library, which only operates on integers.

I would also favour explicitly writing ``std::abs(...) > 0`` to emphasize what is being tested here.

fvm/Equation/DiffusionTerm.set.cpp
```diff
@@ -44,12 +44,15 @@
                 /////////////////////////
                 // RADIUS
                 /////////////////////////
-
                 // Trapping BC: even if the cell is not ignorable, it may still 
                 // be such that the radial flux should be mirrored 
-                if(!isNegativeTrappedRadial && ( Drr(ir, i, j, drr) || Drr(ir+1, i, j, drr) ) ){
+                
+                // For the Ions in DREAM the transient term is defined such that the 
+                // diffusion coefficients should must be negative to get the correct
```
hoppe93, line 51: "Should" or "must"?

fvm/Equation/DiffusionTerm.set.cpp
hoppe93, line 54: I don't quite follow the logic here... To me it seems that the ``abs(...)`` checks if the diffusion coefficient is non-zero, which could equivalently (and more clearly and safely) be done with ``Drr(...) != 0``. I also don't quite understand which situation the comment above refers to? Would it happen in a case where the coefficient provided by the user is still "correct"? Or is this a way of changing the sign of a diffusion coefficient which was given the "wrong" sign by the user?

fvm/Equation/Operator.cpp
```diff
@@ -24,24 +26,48 @@ Operator::~Operator() {}
...
+ * to make it possible to add also advection-diffusion
+ * terms as ordinary terms, ie not via an AdvectionDiffusionTerm.
+ * This is needed for example for advection-diffusion terms on a
+ * quantity with more multiples than one (e.g. ions in DREAM), as the
+ * AdvectionDiffusionTerm class is presently not adapted for such terms.
  */
-void Operator::AddTerm(AdvectionTerm *a) {
-    if (adterm == nullptr)
-        adterm = new AdvectionDiffusionTerm(this->grid);
+void Operator::AddTerm(AdvectionTerm *a, bool addAsEquationTerm) {
```
hoppe93, line 36: Could an equivalent effect be achieved by defining a method ``AddTerm(IonChargedAdvectionDiffusionTerm<AdvectionTerm>*)`` instead of using the new flag  ``addAsEquationTerm``?

fvm/Equation/Operator.cpp
```diff
@@ -24,24 +26,48 @@ Operator::~Operator() {}
...
+ * to make it possible to add also advection-diffusion
+ * terms as ordinary terms, ie not via an AdvectionDiffusionTerm.
+ * This is needed for example for advection-diffusion terms on a
+ * quantity with more multiples than one (e.g. ions in DREAM), as the
+ * AdvectionDiffusionTerm class is presently not adapted for such terms.
  */
-void Operator::AddTerm(DiffusionTerm *d) {
-    if (adterm == nullptr)
-        adterm = new AdvectionDiffusionTerm(this->grid);
+void Operator::AddTerm(DiffusionTerm *d, bool addAsEquationTerm) {
```
hoppe93, line 59: Same comment as for the ``AdvectionTerm`` case...

include/DREAM/Equations/Fluid/IonBoundaryCondition.tcc
```diff
@@ -0,0 +1,145 @@
...
+#include "DREAM/Equations/Fluid/IonBoundaryCondition.hpp"
+
+
+/**
+ * Constructor.
+ */
+template<class T>
+IonBoundaryCondition<T>::IonBoundaryCondition(
+    FVM::Grid *g, IonHandler *ihdl, const len_t iIon,
+    Args&& ... args
```
hoppe93, line 15: Does this compile? I thought some additional template declarations had to be added for this parameter packing to work?

fvm/Equation/DiffusionTerm.set.cpp
vaoskar: Must ;) (unless we have some weird case with effectively negative diffusion coefficients)

fvm/Equation/DiffusionTerm.set.cpp
vaoskar: Well, the `Drr` has to be negative for the ions due to the IonTransientTerm having a different sign than the ordinary TransientTerm, while for other quantities `Drr` should be positive. (the user still gives positive coefficents in the python interface, and then `Drr` is set to minus the user-given value). But as we don't know here what quantity `Drr` applies to, we need to accept both positive and negative `Drr`. And yes, the logic would be the same if the `abs()` were replaced by `Drr(...) != 0`.

fvm/Equation/Operator.cpp
vaoskar: I think so, yes

include/DREAM/Equations/Fluid/IonBoundaryCondition.tcc
vaoskar: I didn't touch the boundary conditions, I thought this was something that you did because you needed it for STREAM? So, you tell me?

include/DREAM/Equations/Fluid/IonChargedAdvectionDiffusionTerm.tcc
```diff
@@ -0,0 +1,114 @@
...
+void IonChargedAdvectionDiffusionTerm<T>::SetCSVectorElements(
+    real_t *vec, const real_t *f, const len_t /*iIon*/, const len_t Z0, const len_t rOffset
+) {
+	if(Z0<1)
+		return;
+		
+    SetCoeffs(Z0); 
+    this->T::SetVectorElements(vec+rOffset, f+rOffset);
+    // Indexing was originally copied from IonKineticIonizationTerm, 
+    // but there the last rOffset was missing. Should it be like that?
```
hoppe93, line 113: Looks right to me! I think it's correct in IonKineticIonizationTerm, because the offset is added later in ``MomentQuantity::SetVectorElements()``. Here, since the ``SetVectorElements()`` is owned by either ``AdvectionTerm`` or ``DiffusionTerm``, the code for ``SetVectorElements()`` assumes that the quantity has size ``nr``. Thus, you should indeed add ``rOffset``, which is defined as ``Zindex*nr``, where ``Zindex`` is the index of the curent charge state.

include/DREAM/Equations/Fluid/IonEquationTerm.tcc
```diff
@@ -21,6 +21,15 @@ IonEquationTerm<T>::IonEquationTerm(FVM::Grid *g, IonHandler *ihdl, const len_t
...
 
+/**
+ * Constructor for AdvectionTerm or DiffusionTerm
+ */
+template<class T>
+IonEquationTerm<T>::IonEquationTerm(FVM::Grid *g, IonHandler *ihdl, const len_t iIon, bool allocCoefficients)
+    : T(g, allocCoefficients), ions(ihdl), iIon(iIon) {
+
+    this->Zion = this->ions->GetZ(iIon);
+}
```
hoppe93, line 32: I would favour just keeping this constructor and give ``allocCoefficients`` a default value in ``IonEquationTerm.hpp``. Then  would avoid having to implement it twice.

include/DREAM/Equations/Fluid/IonEquationTerm.hpp
```diff
@@ -19,6 +19,7 @@ namespace DREAM {
 
     public:
         IonEquationTerm(FVM::Grid*, IonHandler*, const len_t iIon);
+        IonEquationTerm(FVM::Grid*, IonHandler*, const len_t iIon, bool allocCoefficients);
```
hoppe93, line 22: Remove the other constructor and change the last parameter here to ``bool allocCoefficients=false``?

include/DREAM/Equations/Fluid/IonEquationTerm.tcc
hoppe93, line 32: Not sure if it's possible, but if it is, it would be nice to just keep this constructor and give ``allocCoefficients`` a default value.

include/DREAM/Equations/Fluid/IonNeutralAdvectionDiffusionTerm.tcc
```diff
@@ -0,0 +1,67 @@
...
+IonNeutralAdvectionDiffusionTerm<T>::~IonNeutralAdvectionDiffusionTerm(){}
+
+template<class T>
+bool IonNeutralAdvectionDiffusionTerm<T>::SetCSJacobianBlock(
+    const len_t uqtyId, const len_t derivId, FVM::Matrix *jac, const real_t *f,
+    const len_t /*iIon*/, const len_t Z0, const len_t rOffset
+){
+    bool contributes = false;
+    
+    if(!this->HasJacobianContribution(derivId) || Z0>0)
```
hoppe93, line 24: Switch the order of these checks. If statements are always evaluated left -> right, and comparing an integer to a constant is always way faster than calling ``HasJacobianContribution()``.

include/FVM/Equation/Operator.hpp
```diff
@@ -42,8 +42,8 @@ namespace DREAM::FVM {
 
         ~Operator();
 
-        void AddTerm(AdvectionTerm *a);
-        void AddTerm(DiffusionTerm *d);
+        void AddTerm(AdvectionTerm *a, bool addAsEquationTerm = false);
+        void AddTerm(DiffusionTerm *d, bool addAsEquationTerm = false);
```
hoppe93, line 46: (see my comment for the cpp file)

src/Equations/Fluid/IonChargedAdvectionTerm.cpp
```diff
@@ -0,0 +1,100 @@
...
+#include "FVM/Grid/Grid.hpp"
+
+#include <iostream>
+
+using namespace DREAM;
+
+IonChargedAdvectionTerm::IonChargedAdvectionTerm(FVM::Grid *grid, IonHandler *ions, const len_t iIon, 
+    bool allocCoefficients, FVM::AdvectionInterpolationCoefficient::adv_interpolation intp,
+    OptionConstants::adv_jacobian_mode jac_mode, len_t id, real_t damping_factor) : IonChargedAdvectionDiffusionTerm<FVM::AdvectionTerm>(grid, ions, iIon, allocCoefficients){
+	    std::cout<<"check charged 2 \n";
```
hoppe93, line 11: Debug?

src/Equations/Fluid/IonChargedAdvectionTerm.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+    len_t idx = this->ions->GetIndex(iIon, 0);
+    for(len_t Z0 = 1; Z0<=Zion; Z0++,idx++)
+        this->deltars.push_back(new FVM::AdvectionInterpolationCoefficient(grid, FVM::FLUXGRIDTYPE_RADIAL, FVM::AdvectionInterpolationCoefficient::AD_BC_MIRRORED, FVM::AdvectionInterpolationCoefficient::AD_BC_DIRICHLET, idx*nr));
+    
+    this->interpolationCoefficientsShared = false;
+}
+
+void IonChargedAdvectionTerm::DeallocateInterpolationCoefficients(){
+    for (auto it = deltars.begin(); it != deltars.end(); it++)
```
hoppe93, line 32: Just FYI C++ also has a more compact syntax for this:
```
for (auto dt : deltars)
```
In this case, ``dt`` will not be an iterator but a reference to the element directly (i.e. what you get when you dereference ``it`` now, ``*it``)

src/Equations/Fluid/IonChargedDiffusionStochasticBTerm.cpp
```diff
@@ -0,0 +1,110 @@
...
+    for(len_t Z0=1; Z0<=Zion; Z0++)
+        delete [] this->dDrrdNi[Z0-1];
+    delete [] this->dDrrdNi;
+}
+
+void IonChargedDiffusionStochasticBTerm::SetCoeffsAllCS(const real_t t){
+}
+
+void IonChargedDiffusionStochasticBTerm::SetDiffCoeffsAllCS(const real_t t){
+}
```
hoppe93, line 77: What is the purpose of these methods?

hoppe93: Aha, they are declared virtual and null in the parent class? Then this is fine. (If it's not necessary to implement the methods in the derived class, I typically let the parent class implement the methods like this however to not have to write the extra few lines in every child class)

src/Settings/Equations/ions.cpp
```diff
@@ -311,6 +362,81 @@ void SimulationGenerator::ConstructEquation_Ions(EquationSystem *eqsys, Settings
...
+            nZ0_neutral_prescribed_diffusion++;
+        }
+    }
+    
+    // Find number of charge states with prescribed advection
+    len_t nZ0_charged_prescribed_advection = 0;
+    len_t nZ0_neutral_prescribed_advection = 0;
+    for(len_t iZ=0;iZ<nZ;iZ++){
+        cout<<"charged advection modes: "<<charged_advection_mode[iZ]<<"\n";
+        cout<<"neutral advection modes: "<<neutral_advection_mode[iZ]<<"\n";
```
hoppe93, line 383: Debug?

src/Settings/Equations/ions.cpp
```diff
@@ -311,6 +362,81 @@ void SimulationGenerator::ConstructEquation_Ions(EquationSystem *eqsys, Settings
...
+        if(charged_advection_mode[iZ] == OptionConstants::ION_CHARGED_ADVECTION_MODE_PRESCRIBED){
+            nZ0_charged_prescribed_advection += Z[iZ];
+        }
+        if(neutral_advection_mode[iZ] == OptionConstants::ION_NEUTRAL_ADVECTION_MODE_PRESCRIBED){
+            nZ0_neutral_prescribed_advection++;
+        }
+    }
+    
+    cout<<" charged advection: "<<nZ0_charged_prescribed_advection<<"\n";
+    cout<<" neutral advection: "<<nZ0_neutral_prescribed_advection<<"\n";
```
hoppe93, line 393: Debug?

src/Settings/Equations/ions.cpp
```diff
@@ -311,6 +362,81 @@ void SimulationGenerator::ConstructEquation_Ions(EquationSystem *eqsys, Settings
...
+            offsetNeutralDiffusion++;
+        }
+    }
+    
+    // Add advection terms
+    len_t offsetChargedAdvection = 0;
+    len_t offsetNeutralAdvection = 0;
+    for(len_t iZ=0;iZ<nZ;iZ++){
+        if(charged_advection_mode[iZ] == OptionConstants::ION_CHARGED_ADVECTION_MODE_PRESCRIBED){
+            cout<<"check charged \n";
```
hoppe93, line 430: Debug?

src/Settings/Equations/ions.cpp
```diff
@@ -311,6 +362,81 @@ void SimulationGenerator::ConstructEquation_Ions(EquationSystem *eqsys, Settings
...
+            eqn->AddTerm(new IonChargedPrescribedAdvectionTerm(fluidGrid, ih, iZ, true, adv_interp_r_charged, adv_jac_mode_r_charged, id_ni, fluxLimiterDampingCharged, offsetChargedAdvection, FrChargedPrescribed), true);
+            offsetChargedAdvection+=Z[iZ];
+        }
+        if(neutral_advection_mode[iZ] == OptionConstants::ION_NEUTRAL_ADVECTION_MODE_PRESCRIBED){
+            cout<<"check neutral \n";
```
hoppe93, line 435: Debug?

py/DREAM/Settings/Equations/IonSpecies.py
```diff
@@ -447,6 +614,187 @@ def initialize_prescribed_charge_state(self, Z0, n=None, r=None, t=None, interpr
             self.initialize_prescribed(n=N, t=t, r=r)
         else:
             raise EquationException("ion_species: '{}': Unrecognized shape of prescribed density: {}.".format(self.name, n.shape))
+            
+            
+            
+            
+    def initialize_charged_prescribed_diffusion(self, charged_prescribed_diffusion=None, rChargedPrescribedDiffusion=None, tChargedPrescribedDiffusion=None, interpr=None, interpt=None):
```
hoppe93, line 621: General comment about all these ``initialize`` routines: do you think it would be possible to put most of the logic in a general initialization routine which utilizes Python's dynamic variables to avoid duplicating code? I imagine that these initialization methods would remain, but their contents would be reduced to calling a general method which also takes an additional parameter which indicates the name of the relevant variables to set. Instead of setting ``self.tChargedPrescribedDiffusion`` explicitly, one would then call ``setattr(self, f"t{name}PrescribedDiffusion", value)``, or something like that.

(This is how it's done in the general transport interface, ``py/DREAM/Settings/Transport.py``)

py/DREAM/Settings/Equations/Ions.py
```diff
@@ -2,13 +2,29 @@
...
+AD_INTERP_DOWNWIND = AdvectionInterpolation.AD_INTERP_DOWNWIND
+AD_INTERP_QUICK    = AdvectionInterpolation.AD_INTERP_QUICK 
+AD_INTERP_SMART    = AdvectionInterpolation.AD_INTERP_SMART 
+AD_INTERP_MUSCL    = AdvectionInterpolation.AD_INTERP_MUSCL 
+AD_INTERP_OSPRE    = AdvectionInterpolation.AD_INTERP_OSPRE 
+AD_INTERP_TCDF     = AdvectionInterpolation.AD_INTERP_TCDF  
+
+AD_INTERP_JACOBIAN_LINEAR = AdvectionInterpolation.AD_INTERP_JACOBIAN_LINEAR
+AD_INTERP_JACOBIAN_FULL   = AdvectionInterpolation.AD_INTERP_JACOBIAN_FULL  
+AD_INTERP_JACOBIAN_UPWIND = AdvectionInterpolation.AD_INTERP_JACOBIAN_UPWIND
```
hoppe93, line 26: This is fine, but I just want to mention that the reason we have declarations like these in f_hot and f_re is because of backwards compatibility. When the advection interpolation classes were originally implemented, we only applied them to distribution functions. Later, when we added support for advection interpolation in the transport interface, we added the common ``AdvectionInterpolation`` class in the Python interface. I now recommend only taking the settings for the interpolation coefficients from that class, but since a fair number of scripts used the old interface with constants defined in ``HotElectronDistributionFunction.py`` we defined coefficients there as well.

So in short, in my view of how we should write scripts in the future, these definitions are unnecessary. As long as they map to the same constants in ``AdvectionInterpolation``, this is fine though.

fvm/Equation/DiffusionTerm.set.cpp
hoppe93: Ok... But so it doesn't really have to with the sign of ``Drr``, but just whether it's non-zero or not?

When taking the absolute value in C++, you should make a habit of explicitly writing ``std::abs()`` since you might otherwise get the C version of ``abs()`` without the compiler telling you. The C version only operates on integers, so in this case any diffusion coefficient with ``|Drr| < 1`` would be treated as identically zero (assuming you are unlucky and get the C version of ``abs()``). I would still highly recommend checking for zero explicitly instead, since it's both clearer and faster.

include/DREAM/Equations/Fluid/IonBoundaryCondition.tcc
hoppe93: Then it probably compiles and there's just some magic which I can't see now going on :)

py/DREAM/Output/FluidQuantity.py
```diff
@@ -152,7 +152,7 @@ def get(self, r=None, t=None):
             return self.data[t,r]
 
         
-    def plot(self, ax=None, show=None, r=None, t=None, log=False, colorbar=True, VpVol=False, weight=None, **kwargs):
+    def plot(self, ax=None, show=None, r=None, t=None, log=False, colorbar=True, VpVol=False, weight=None, ms_time=False, **kwargs):
```
hoppe93, line 155: Ooh, this looks like a good idea! Maybe we could even generalize it further so that it's a string indicating if we want seconds, milliseconds, microseconds or nanoseconds? I can do that.

py/DREAM/Output/IonSpecies.py
```diff
@@ -138,4 +139,23 @@ def plot(self, t=None, r=None, Z0=None, ax=None, show=None):
...
+        
+    def plotSum(self, Z0 = None, integrate = False, **kwargs):
+        if Z0 is None: Z0 = slice(None)
+        states = self.ionstates[Z0]
+        data = None
+        for state in states:
+            if data is None:
+                data = state.data
+            else:
+                data = data + state.data
```
hoppe93, line 151: Or perhaps
```python
data = np.zeros(states[0].data.shape)
for state in states:
    data += state.data
```
? (no need to change; just wanted to point this out :)

src/Equations/Fluid/KineticEquationTermIntegratedOverMomentum.cpp
```diff
@@ -30,6 +30,10 @@ KineticEquationTermIntegratedOverMomentum::KineticEquationTermIntegratedOverMome
     rebuildOperator(rebuild), scaleFactor(scaleFactor) {
     
     SetName("KineticEquationTermIntegratedOverMomentum");
+    
+    this->CsetJacobian = new Mat[u->GetNUnknowns()];
```
hoppe93, line 34: This should also be deleted in the destructor, right?

vaoskar: Isn't it already (line 48-49)?

py/DREAM/Output/IonSpecies.py
```diff
@@ -138,4 +139,23 @@ def plot(self, t=None, r=None, Z0=None, ax=None, show=None):
             plt.show(block=False)
 
         return ax
-
+        
+    def plotSum(self, Z0 = None, integrate = False, **kwargs):
```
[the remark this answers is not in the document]
vaoskar: Done!

src/Equations/Fluid/KineticEquationTermIntegratedOverMomentum.cpp
hoppe93: No, I meant deleting the pointer:
```c++
delete [] this->CsetJacobian;
```

hoppe93: What is on lines 48-49 now is also needed, but it doesn't undo the call to ``new`` on line 34.

vaoskar: Ah, right, I updated it now!

hoppe93 (approved):
Looks good now! Well done! :)

Issue #147, opened 2021-10-29
Debugging information for J, F, and dx are out of sync in SolverNonLinear.cpp


aejarvin:
In the SolverNonLinear.cpp (around lines 325 ->), the SaveDebugInfo is called after updating F and J, but before J*dx = F is solved. As a result, the printed debugging dx is out of sync with F and J. The issue can be solved by adding a dedicated DebugInfo-writing routine for dx after solving J*dx = F.

PR #148, opened 2021-11-01
Implement fix for #147

hoppe93:
Make sure that the new solution vector "dx" obtained as debug output with the non-linear solver corresponds to the actual solution of the specified timestep/iteration, and not of the previous timestep/iteration. The old "SaveDebugInfo()" method has been divided into two separate methods which should be called before and after "Invert()" is called, respectively. This allows the residual and Jacobian to be evaluated and saved *before* inversion (in case the code crashes during inversion), while the new solution is evaluated after inversion.

Fix implemented in response to issue #147.


vaoskar (approved):
Looks good, can not find anything to add

PR #154, opened 2021-12-03
Implemented the analytical avalanche growth rate (eq. 18 in Rosebluth…

PajterMclovin:
… & Putvinski 1997)

### DIFF
--- a/src/Equations/RunawayFluid.cpp
+++ b/src/Equations/RunawayFluid.cpp
@@ -39,13 +39,14 @@ RunawayFluid::RunawayFluid(
     OptionConstants::eqterm_dreicer_mode dreicer_mode,
     OptionConstants::collqty_Eceff_mode Eceff_mode,
     OptionConstants::eqterm_avalanche_mode ava_mode,
+    OptionConstants::eqterm_avalanche_fluid_factor ava_factor,
     OptionConstants::eqterm_compton_mode compton_mode,
     real_t compton_photon_flux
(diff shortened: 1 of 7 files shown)


src/Equations/RunawayFluid.cpp
```diff
@@ -336,7 +337,16 @@ void RunawayFluid::CalculateGrowthRates(){
 
     for (len_t ir = 0; ir<this->nr; ir++){
         avalancheGrowthRate[ir] = n_tot[ir] * constPreFactor * criticalREMomentumInvSq[ir];
-        real_t pc = criticalREMomentum[ir]; 
+
+        // multiply by trapping correction factor
+        if (ava_trapping == OptionConstants::EQTERM_AVALANCHE_TRAPPING_CORRECTION_INCLUDE){
+
+            real_t eps = (this->rGrid->GetR(ir)) / (this->rGrid->GetR0());
+            real_t phi = 1/(1 + 1.46*sqrt(eps) + 1.72*eps);
```
Embreus, line 345: isn't this phi just a poor man's approximation to our rGrid->effectivePassingFraction which is the correct generalization to arbitrary geometry? @hoppe93

Embreus:
I think it's really neat to implement RP's equation 18. Based on the phrasing of the setttings it sounds like the previously implemented model doesn't account for trapping... but I thought it did? Shouldn't you, in a pure plasma, get effectively the sqrt(phi) scaling with radius using what was already in DREAM? 

disclaimer: high risk of my memory being foggy, and I didn't go digging too deeply.

PR #160, opened 2022-02-09
Iss159

vaoskar:


tools/eqget/GEQDSK.py
```diff
@@ -38,18 +38,30 @@ def _next_value(self, fh):
...
+        #else:
+        #    vertices = v[-1]
+            
+        vertices = self.contour_generator.create_contour(psi_n)
         
-        R, Z = vertices[0][:,0], vertices[0][:,1]
+        iClosedContour = None
+        for i in range(len(vertices)):
+            if len(vertices[i].shape)==2:
+                if np.sqrt((vertices[i][0,0]-vertices[i][-1,0])**2 + (vertices[i][0,1]-vertices[i][-1,1])**2)<closedContourTol:
```
hoppe93, line 57: Is this necessary? I haven't checked myself, but I thought ``create_contour()`` would set the first and last points to exactly the same...

hoppe93: (this = introducing a tolerance for the comparison)

vaoskar: I just tried checking if the are exactly the same, ie changing line 57 to

`if vertices[i][0,0]==vertices[i][-1,0] and vertices[i][0,1]==vertices[i][-1,1]:`

and it didn't work. So, I guess, not really (and isn't it always a bit scary to depend on floats being exactly the same)?

hoppe93: Ok! Well, I thought the routine ``create_contour()`` would explicitly copy the first values to the last element when the contour is closed, since that would guarantee an actually closed contour. But I suppose it's difficult to tell if the contour is actually closed... This will do fine then!

hoppe93 (approved):
Excellent!

PR #165, opened 2022-02-16
Fix forgotten dependence of psi_p on psi_edge

hoppe93:
Apparently I had forgotten to make ``psi_p`` depend on ``psi_edge`` when the former is initialized by solving Ampère's law. In that case, ``psi_edge`` could be initialized *after* ``psi_p``, so that the values of ``psi_edge`` and ``psi_p`` would be inconsistent.


EsmeeBerger (approved):
With these changes the issue seems to be resolved, allowing DREAM-simulations with non-linear initialization to run without crashing even when rWall is not equal to a.

PR #166, opened 2022-02-18
Hydrogen support

hoppe93:
For STREAM we needed to differentiate between H, D and T. On this branch we therefore implement support for that by introducing a new flag ``hydrogen`` in the Python interface and an implementation which is analogous to the one for tritium.

### DIFF
--- a/include/DREAM/Constants.hpp
+++ b/include/DREAM/Constants.hpp
@@ -17,6 +17,9 @@ namespace DREAM {
         //Elementary charge in C:
         static const real_t ec;
 
+		//Hydrogen mass in kg
+		static const real_t mH;
+
         //Deuterium mass in kg
         static const real_t mD; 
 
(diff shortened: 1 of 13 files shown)


vaoskar (commented):
Looks good, but I could perhaps mention that I implemented the ability to specify an isotope-number for any ion species as this information is needed in the SPIHandler. Currently I think this property is only used by the SPIHandler, but it would perhaps be more consistent to use it to also differ between the hydrogen isotopes in other clases? But perhaps you don't want to bother with this now? In that case we could just merge this pull request and just create an issue about this instead.

hoppe93:
Ok, having the possibility to specify isotope number would be very nice, but it seems to me that the isotope functionality is only part of the ``SPIHandler`` and does not affect the ion mass in ``IonHandler``? Since we have not included SPI functionality in STREAM it would probably require quite a bit of work to get a more generally applicable isotope functionality, and so it's probably better to just merge this now?

vaoskar:
Wel, you wouldn't have to implement SPI functionality to make it work, but you are right that I haven't implemented any use of the isotope number in the IonHandler yet, so it seems reasonable to merge this now to not delay your work.

vaoskar (approved):

PR #171, opened 2022-03-17
CSD3 setup scripts

afil-ukaea:
Environment and build scripts for the CSD3 cluster (Cambridge, UK). I had to use a conda environment to be able to install all required python packages on the cluster. I have tested that the code is running correctly with testDREAM.sh as well as examples/basic and Disruptions.


hoppe93 (approved):
Great! Thanks for adding this. Feel free to merge whenever the tests finish :)

PR #178, opened 2022-06-17
Adaptive time stepper based on ionization time

hoppe93:
This pull request implements an adaptive time stepper which estimates the instantaneous ionization time scale and rescales the time step accordingly. The user specifies the initial time step ``dt0``, which is used for the very first time step. After the first time step has been completed, the ionization time scale can be evaluated for the first time, and this time scale, ``timescale0`` is used as the baseline. In subsequent, the ionization time scale ``timescale`` is evaluated, and the time step is estimated according to
```
dt = dt0 * timescale / timescale0
```
This has been tested on a TCV disruption case (in fluid mode), but will also be tested on an isotropic+kinetic RE simulation before being merged.

Documentation will also be added before final merge.

### DIFF
--- a/include/DREAM/Settings/OptionConstants.enum.hpp
+++ b/include/DREAM/Settings/OptionConstants.enum.hpp
@@ -138,7 +138,8 @@ enum linear_solver {
 /////////////////////////////////////
 enum timestepper_type {
     TIMESTEPPER_TYPE_CONSTANT=1,
-    TIMESTEPPER_TYPE_ADAPTIVE=2
+    TIMESTEPPER_TYPE_ADAPTIVE=2,
+	TIMESTEPPER_TYPE_IONIZATION=3
 };
 
 /////////////////////////////////////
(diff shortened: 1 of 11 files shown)


hoppe93:
Before merging I would also like to add a feature which automatically estimates the appropriate baseline time step ``dt0`` to use. I imagine that this will done in a few iterations at the start of the simulation:
- Take a very short time step which should not fail in any realistic scenario (e.g. ``dt=1e-12``). Based on the result, and a user-specified scaling parameter (which will be set to 50 by default, which seems to be a good value, at least in the isotropic+kinetic RE simulation; lower values may be desirable in fluid simulations and so we might want to let the C++ kernel determine the default value), we estimate the baseline time step ``dt0`` as ``dt0=timescale/scaleparameter``.
- If the simulation crashes with the selected ``dt0``, we can automatically restart the time stepping with a halved value for ``dt0``. This should only be repeated a limited number of times (e.g. 3 times), after which the solver should conclude that the simulation is not going to converge anyway.

hoppe93:
First tests in the isotropic+kinetic RE mode were successful in the sense that the solver seems to be very stable and reduced the simulation time from ~20 hours to ~2 hours. The result however differs significantly from the simulation using two stages of the constant time stepper.

hoppe93:
I also implemented a feature to automatically adjust ``dt0`` at the start of the simulation by taking a short step ``dt=1e-12``, after which the ionization time scale is estimated and ``dt0`` adjusted accordingly based on a "safetyfactor" parameter. This however did not work as well as expected since it seems that the required safety factor can vary significantly for different types of simulations. Therefore, the user would anyway have to carefully tweak the ``safetyfactor`` parameter, and I think it might then be better for the user to directly set the ``dt0`` parameter which is more intuitive.

PR #182, opened 2022-08-25
minor bug fix

peterhalldestam:
Changed environment variable DREAM_DIR to DREAMPATH to be consistent with what is set in environment.tok.sh.


hoppe93 (approved):
Perfect!

PR #184, opened 2022-09-14
Iontransport

vaoskar:
This pull request includes various features allowing for more flexibility in setting up SPI simulations, and some helper functions to more easily set transport coefficients in a few ways used within our collaboration with ITER. Some of it has not been tested thoroughly, but shouldn't hurt anyone even if further development might be desirable.


include/DREAM/Equations/Scalar/ConstantSPIVelocityPositionTerm.hpp
```diff
@@ -46,12 +50,29 @@ namespace DREAM {
                 this->initial_data=xp_init;
             }
         }
+        void SetTimeDelay(const real_t *t_delay, bool copy=true){
```
hoppe93, line 53: This file seems to be containing more method implementations than other declarations. It would be better to separate it into a ``.hpp`` containing declarations and a ``.cpp`` containing the method implementations (constructor, destructor, ``GetData()``, ``SetTimeDelay()`` and ``Rebuild()``).

include/DREAM/Equations/Scalar/ConstantSPIVelocityPositionTerm.hpp
```diff
@@ -46,12 +50,29 @@ namespace DREAM {
...
+            if(copy){
+                real_t *data_tmp=new real_t[nShard];
```
hoppe93, line 55: I cannot find any corresponding ``delete`` for this one. And I hope that the input ``t_delay`` is deleted by the caller?

py/DREAM/Settings/Equations/IonSpecies.py
```diff
@@ -228,6 +228,16 @@ def setTemperature(self, T):
              raise EquationException("ion_species: '{}': Invalid dimensions of initial ion temperature T: {}x{}. Expected {}x{}."
                 .format(self.name, T.shape[0], T.shape[1], 1, np.size(self.r)))        
         return T
+        
+    def shiftTimeTranspCoeffs(self, tShift):
```
hoppe93, line 232: Please add a comment to this method describing what it's purpose is.

py/DREAM/Settings/Equations/Ions.py
```diff
@@ -311,6 +311,17 @@ def setIonType(self, index, ttype):
             raise DREAMException("Trying to set invalid ion type for ion species '{}': {}.".format(ion.name, ttype))
 
         ion.ttype = ttype
+
+        
+    def shiftTimeTranspCoeffs(self, tShift):
```
hoppe93, line 316: Please add a comment to this method describing what the method is supposed to do.

tools/ITER_Profile_reader.py
```diff
@@ -0,0 +1,66 @@
+# This class provides routines for reading input profiles from CORSICA data files
+
+import numpy as np
+
+class ITER_Profile_reader:
```
hoppe93, line 5: If this script loads CORSICA profiles, wouldn't a more natural name be ``CORSICA_Profile_reader``? (Or even ``CORSICAProfileReader``?)

tools/ITER_Profile_reader.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def __init__(self, filename):
+        self.filename = filename
+        
+        f = open(self.filename,'r')
+        lines = f.readlines()
```
hoppe93, line 11: The proper way of reading a text file in Python is as follows:
```python
with open(self.filename, 'r') as f:
    lines = f.readlines()
```
The ``with`` ensures that the file is closed when reading is finished.

tools/ITER_Profile_reader.py
```diff
@@ -0,0 +1,66 @@
...
+        keep_reading = True
+        
+        # Go through the file and find basic scalar data 
+        # (number of columns ncol for the profile datasets, 
+        # number of radial gridpoints N_Rho and the plasma current Ip)
+        while keep_reading:
+            line = lines[iLine]
+            #print(line)
+            if line.split():
+                if line.split()[0] == 'ncol':
```
hoppe93, line 22: ``line.split()`` creates an array. Wouldn't it better to store that array once in each ``while`` iteration, instead of calling ``line.split()`` multiple times?

tools/ITER_Profile_reader.py
```diff
@@ -0,0 +1,66 @@
+# This class provides routines for reading input profiles from CORSICA data files
+
+import numpy as np
+
+class ITER_Profile_reader:
+
+    def __init__(self, filename):
+        self.filename = filename
+        
+        f = open(self.filename,'r')
```
hoppe93, line 10: I would recommend putting the remainder of the constructor in a separate method which could be called again if necessary.

tools/ITER_Profile_reader.py
```diff
@@ -0,0 +1,66 @@
...
+            iLine+=1        
+            keep_reading = iLine<len(lines)
+        
+    def load_profile(self, quantity):
+        # This function goes through the file and finds the profile dataset named 'quantity' 
+        # in the input file, where 'quantity' should be a string matching one of the 
+        # quantity names in the CORSICA file, and returns this profile as a numpy array
+        
+        f = open(self.filename,'r')
+        lines = f.readlines()
```
hoppe93, line 43: Use ``with`` instead (see above)

tools/ITER_Profile_reader.py
```diff
@@ -0,0 +1,66 @@
...
+                    self.N_Rho = int(lines[iLine+1])
+                    iLine+=1
+                    
+                if line.split()[0] == 'plasma' and line.split()[1] == 'current':
+                    self.Ip = float(lines[iLine+1].replace('D','e'))
+                    iLine+=1
+                            
+            
+            iLine+=1        
+            keep_reading = iLine<len(lines)
```
hoppe93, line 35: It seems this is the only line modifying ``keep_reading``? If so, you could consider putting this boolean expression directly in the loop head instead.

tools/ITER_Profile_reader.py
```diff
@@ -0,0 +1,66 @@
...
+                    while iRho<self.N_Rho:
+                        line = lines[iLine+1]
+                        for iCol in range(len(line.split())):
+                            values[iRho] = float(line.split()[iCol].replace('D','e'))
+                            iRho+=1
+                        iLine+=1
+                            
+            
+            iLine+=1        
+            keep_reading = iLine<len(lines)
```
hoppe93, line 61: Put the boolean expression in the loop head instead?

tools/eqget/GEQDSK.py
```diff
@@ -50,6 +50,7 @@ def get_flux_surface(self, psi_n, theta=None, closedContourTol = 1e-6):
         #    vertices = v[-1]
             
         vertices = self.contour_generator.create_contour(psi_n)
+        vertices = vertices[0]
```
hoppe93, line 53: This reminds me... This line is due to a change made in some version of matplotlib. We should have an if statement instead since this will break on older versions. I'll add this on a separate branch, so just ignore this comment for now.

tools/transport_coeffs_reader.py
```diff
@@ -0,0 +1,247 @@
+# This class provides routines for setting the electron transport coefficients (heat transport and runaway transport),
+# either by reading the from a file or by estimating them with some of the available models. 
+# Some functions in this class are modified versions of previous work done by Andreas Sundström
```
hoppe93, line 3: Andréas's name is misspelled...

tools/transport_coeffs_reader.py
```diff
@@ -0,0 +1,247 @@
...
+import DREAM.Settings.TransportSettings as Transport
+
+## Elementary charge
+c      = 299792458.      # m/s
+e      = 1.602176634e-19 # C
+m_e    = 9.1093837e-31   # kg
+m_e_eV = 510.999e3       # eV
+
+
+class transport_coeffs_reader:
```
hoppe93, line 24: ``TransportCoefficientReader`` would be more consistent with all other scripts.

hoppe93 (changes requested):
Looks mostly good, although I have some comments in a few places.

py/DREAM/Settings/Equations/Ions.py
vaoskar: It is similar as with the ion densities, you specify the transport coefficients separately for each species, but the grids on which you specify them must be the same for all species.

tools/transport_coeffs_reader.py
```diff
@@ -0,0 +1,247 @@
+# This class provides routines for setting the electron transport coefficients (heat transport and runaway transport),
+# either by reading the from a file or by estimating them with some of the available models. 
+# Some functions in this class are modified versions of previous work done by Andreas Sundström
+
+import numpy as np
+import sys
+import h5py as h5py
+import os as os
+
+sys.path.append('../../py/')
```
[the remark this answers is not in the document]
vaoskar: There is a comment in cli.py about a TODO to set up the path properly there, is that still relevant and does it mean I shouldn't take after how it is done there?

hoppe93 (approved):
Great! This is (finally!) ready to merge as far as I'm concerned.

PR #193, opened 2022-10-13
Limit number of save steps in adaptive ioniz timestepper

hoppe93:
Limit number of save steps in adaptive ioniz timestepper


peterhalldestam (approved):
Perfect, works great!

PR #196, opened 2022-11-04
Magnetic compression operator

hoppe93:
This pull request implements an advection operator for the kinetic equation representing the conservation of magnetic moment when the magnetic field strength is varied in time. The new functionality does *not* implement any time variation of the magnetic field, but the operator can be used to approximately study the effect of the magnetic moment conservation. 

(The implementation of this operator was motivated by experiments at TCV in which the magnetic field strength was decreased in order to affect the interaction between runaway electrons and the toroidal field ripple)


peterhalldestam (approved):
Looks good, as far as I can tell!

PR #198, opened 2022-11-11
Implemented terms in ion rate equation into other.fluid

IdaEkmark:


hoppe93 (approved):
Looks good! I just made some changes to where the equation terms are stored (in the ``OtherQuantityHandler`` struct instead of in ``EquationSystem``) to be more consistent with other parts of the code.

PR #199, opened 2022-11-14
Ion source

hoppe93:
An ion source term implemented as a boundary condition for the ion equations.


peterhalldestam:
Question: this implemented source term seems to be able to add particles locally at the plasma edge, but not anywhere else. Would it make sense to include a source term S = S(t, r) that can be add particles inside of the plasma as well as at the edge? Would this ability even be relevant for DREAM?

PR #205, opened 2023-02-01
Updated AMJUEL floating point error

IdaEkmark:


hoppe93:
@vaoskar We had some issues with the AMJUEL functions when running in isotropic mode. Since ``T_cold`` starts from 1eV (in our simulations) in isotropic mode, ``ln(T_cold)`` in the Lyman derivatives became zero which caused division-by-zero and subsequent nan errors. I'm not exactly sure about how the derivatives are to be written, but based on the implementation we took a guess at how it would generalize to T_cold=1. Do you think it makes sense?

vaoskar (commented):
This seems to work (the coefficients are given such that the logarithm of the coefficient is a polynomial in ln(T) and ln(n), which you seem to have interpreted correctly). I do however have two comments:

- It might be marginally more efficient to have a separate set of loops for the case lnT==0, so you don't have to check if lnTi==0 ond i==1 for every i even if only i==1 gives a contribution to derivFactor.
- One should perhaps do a similar fix for the density derivatives. In fact, I looked up the manual again and apparently the fit is only valid for densities between 10^{14} m^{-3} and 10^{22} m^{-3}, and outside this range one should use the closest of these values (the coefficients thay give coincide with the coronal and LTE limit, respectively). You could perhaps fix this as well while you are at it (if I don't find the time to fix it myself first)

hoppe93 (approved):
Great!

PR #217, opened 2023-02-27
Test for the Ampere-Faraday law

hoppe93:
This pull request adds another physics test for the Ampère-Faraday law. The test is inspired by the one used in [Harvey et al, Nuclear Fusion 59 (2019)] and computes the conductivity based on the decay time of the electric field. The theory behind the test is presented in ``doc/notes/pdf/amperefaraday.pdf``.


hoppe93:
Superb! Thanks :D

PR #222, opened 2023-04-13
Parallel runiface

Pinfar:
I created parallel version of runiface. It can handle any number of concurrent jobs.

### DIFF
--- a/py/DREAM/DREAMTask.py
+++ b/py/DREAM/DREAMTask.py
@@ -2,6 +2,7 @@ import os
 import pathlib
 import subprocess
 import tempfile
+import time
 
 from . DREAMException import DREAMException
 from . DREAMOutput import DREAMOutput
@@ -32,9 +33,10 @@ class DREAMTask:
         self.stderr_data = None
         self.DREAMPATH = DREAMPATH
         self.quiet = quiet
-        self.timeout = timeout
+        self.timeout = timeout if timeout is not None else float('inf')
 
     def run(self):
+        self.startTime= time.time()
         if self.p != None: #if process is already created then we can safely od nothing
             return 
         if self.quiet:
@@ -59,9 +61,12 @@ class DREAMTask:
             return False
         except KeyboardInterrupt:
             self.errorOnExit = 2
-        # We still need to implement timeout manually
-        # self.p.kill()
-        # errorOnExit = 3
+
+        # Now we implement timeout manually, by checking start time with current time
+        if time.time() - self.startTime > self.timeout:
+            self.p.kill()
+            self.errorOnExit = 3
+ 
         return True
 
     def getResult(self):
--- a/py/DREAM/DREAMTask.py
+++ b/py/DREAM/DREAMTask.py
@@ -7,6 +7,7 @@ import time
 from . DREAMException import DREAMException
 from . DREAMOutput import DREAMOutput
 from . DREAMSettings import DREAMSettings
+from . DREAMErrorOutput import DREAMErrorOutput
 from subprocess import TimeoutExpired
 
 class DREAMTask:
@@ -81,4 +82,11 @@ class DREAMTask:
             raise DREAMException("DREAMi simulation was killed due to timeout.")
         else:
             return self.obj
-        
+
+    def getResultObject(self):
+        # Handling errors in this way is easiest in terms of backwards compatibility, but
+        # quite inefficient. We can change it later, if we face performance issues.
+        try:
+            return self.getResult()
+        except DREAMException as error:
+            return DREAMErrorOutput(error, self.stderr_data, self.errorOnExit)
--- a/py/DREAM/DREAMTask.py
+++ b/py/DREAM/DREAMTask.py
@@ -1,0 +1,79 @@
+import os
+import pathlib
+import subprocess
+import tempfile
+
+from . DREAMException import DREAMException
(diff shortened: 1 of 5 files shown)


py/DREAM/DREAMTask.py
```diff
@@ -0,0 +1,92 @@
...
+            # In this case it is expected situation. Process has not completed during specified timeout.
+            # We don't need to preserve any state, because documentation says that no stream data will be lost.
+            return False
+        except KeyboardInterrupt:
+            self.errorOnExit = 2
+
+        # Now we implement timeout manually, by checking start time with current time
+        if time.time() - self.startTime > self.timeout:
+            self.p.kill()
+            self.errorOnExit = 3
```
hoppe93, line 69: Should this be in the ``except TimeoutExpired`` statement? It seems to me that either (i) the call to ``self.p.communicate()`` would time out and throw a ``TimeoutExpired`` exception, after which this method would now immediately return ``False``, or (ii) the process would complete before the timeout after which ``self.p.kill()`` might be called (if the ``if`` condition is ``True``) on a process which is already finished?

hoppe93: The reason the timeout test passes now seems to be that ``self.timeout`` is shorter than the runtime of the simulation, so that ``errorOnExit=3`` after the process is finished. But it is not actually killed before the simulation completes.

tests/physics/runiface_parallel/runiface_parallel.py
```diff
@@ -0,0 +1,128 @@
...
+    E = 6       # Electric field strength (V/m)
+    n = 5e19    # Electron density (m^-3)
+    T = 100     # Temperature (eV)
+
+    # Grid parameters
+    pMax = 1    # maximum momentum in units of m_e*c
+    Np   = 300  # number of momentum grid points
+    Nxi  = 20   # number of pitch grid points
+    tMax = 1e-3 # simulation time in seconds
+    Nt   = 20   # number of time steps
```
hoppe93, line 37: You could increase ``Nt`` to something like 200 to add some safety margin to the timeout test. At the moment, on my laptop (with compiler optimizations disabled), each simulation runs in about 0.7s while the timeout is 0.5s. On a faster computer, with compiler optimization, the simulation might finish before the timeout expires.

Actually, maybe it would be best to add ``Nt`` as an input parameter to ``createExperimentData()`` with default value 20. Then, in ``check_timeout()`` you can pass ``Nt=200`` instead just be sure that it will run for long enough.

hoppe93 (changes requested):
Great work! I think there's just a small logical error in ``DREAMTask.hasFinished()``, and I added a suggestion for changing one of the input parameters in the test. If you fix this I'm more than happy to approve and merge to master!

py/DREAM/DREAMTask.py
Pinfar: According to https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate TimeoutExpired exception is thrown when timeout of communication method is expired. This is expected, because we want to communicate in round-robin style with all tasks to see if they have completed/crashed. This is NOT a timeout for whole task! The algorithm here is:
1. run N processes.
2. communicate with each of them to see if it completed. Communicate with short timeout. If it throws TimeoutExpired, then assume that the process is still running and just go to checking next process. If process completes/crashes, start next process from input queue.
3. Repeat until input queue is empty.

tests/physics/runiface_parallel/runiface_parallel.py
Pinfar: Good idea, I will add it in a minute.

Pinfar: Please, check it now.

py/DREAM/DREAMTask.py
hoppe93: I agree with how ``TimeoutExpired`` works, but the placement of this if statement currently makes ``self.timeout`` have a different effect from what we want it to have. This ``if`` statement will only ever be reached if the process has finished *normally*. If the process took longer than ``self.timeout`` to complete, then ``p.kill()`` will be called after the process completed (on a process which is no longer active) and the task will be indicated as failed. The value of ``self.timeout`` will not however limit for how long the process is allowed to run.

The behaviour we want is instead for ``self.timeout`` to limit for how long a process is allowed to run (sometimes simulations hangs, and so if they run for more than, say 30 minutes, we can consider them failed). For this to happen, this if statement should instead go inside the ``except TimeoutExpired`` block.

(I ran the test and, after inserting som ``print``s, could confirm that this is indeed how the code behaves presently)

tests/physics/runiface_parallel/runiface_parallel.py
hoppe93: The update looks good!

py/DREAM/DREAMTask.py
Pinfar: OH, now I see it. You are right obviously :) I'll fix it in the evening. Thanks :)

Pinfar: I fixed it and pushed changes.

hoppe93 (approved):
Looks good! As far as I'm concerned, this is ready to merge!

Pinfar:
I think I am not able to merge PR because of 'The base branch restricts merging to authorized users.' Can you do this yourself?

PR #225, opened 2023-04-20
Stopping power for Tungsten

jjwalkowiak:
I created new branch and put here all changes, as previous branch was broken


hoppe93:
The error in the ``trapping_conductivity`` test is due to that ``atomicParameter`` in ``src/Equations/SlowingDownFrequency.cpp`` is zero (or *just* above zero) for neutral deuterium. Specifically, the code crashes on this line: https://github.com/chalmersplasmatheory/DREAM/blob/bf6ee6d624b4369a442660fabf3a63e904746952/src/Equations/SlowingDownFrequency.cpp#L108.

To debug this, you can save the settings from the simulation by uncommenting this line: https://github.com/chalmersplasmatheory/DREAM/blob/bf6ee6d624b4369a442660fabf3a63e904746952/tests/physics/trapping_conductivity/trapping_conductivity.py#L117. Then run the test manually by going to ``tests/physics`` and invoking ``./runtests.py trapping_conductivity``.

You will then get a file called ``settings_trapping_conductivity.h5`` in ``tests/physics/``. You can run DREAM on this file by calling ``../../build/iface/dreami settings_trapping_conductivity.py``. From this you should be able to debug the problem as usual (e.g. by putting printfs in your code, or using other techniques).

jjwalkowiak:
For me it seems that this test is not using the value for neutral deuterium, but for Z=300. To be honest, I am not sure why it crashes at Z = Z0 and not earlier. All previous values seems to be random numbers.
I added a condition which switch to previous extrapolation for Z>86.

src/Equations/SlowingDownFrequency.cpp
```diff
@@ -124,20 +135,26 @@ real_t SlowingDownFrequency::GetAtomicParameter(len_t iz, len_t Z0){
...
-            S_N0 = Ne - sqrt(Ne*HIGH_Z_EXCITATION_ENERGY_PER_Z / HYDROGEN_MEAN_EXCITATION_ENERGY); // S_N0: for a neutral atom with Z=N
-        }
-        real_t A_N = (1-D_N) * (1-D_N);
-        real_t B_N = 2*(1-D_N) * (Ne*D_N - S_N0);
-        real_t C_N = (Ne*D_N - S_N0) * (Ne*D_N - S_N0);
-
-        I = HYDROGEN_MEAN_EXCITATION_ENERGY * (A_N*Z*Z + B_N*Z + C_N);
+    }
+    else    { /* use extended data from MeanExcitationEnergy_Extended.cpp */
+		if (Z <= MAX_Z_EXTENDED){ /*this if can be used to switch to the old version of code*/
```
hoppe93, line 140: This could be written as
```c++
else if (Z <= MAX_Z_EXTENDED) {
```
instead, to avoid unnecessary levels of nested if's.

tools/Generate_MeanExcitationEnergy.py
```diff
@@ -0,0 +1,125 @@
...
+# #################################################################################
+
+import numpy as np
+import scipy.integrate as integrate
+import math
+import sys
+import pathlib
+import argparse
+
+def GenerateMEE(Z_min=19, Z_max=86):
```
hoppe93, line 20: Should there be a check at the start of this function to ensure that ``Z_min`` and ``Z_max`` are valid, perhaps?

hoppe93 (approved):
Looks good! I added two suggestions for improvements, but they should not be absolutely mandatory for a merge.

tools/Generate_MeanExcitationEnergy.py
jjwalkowiak: I think for now it can stay as it is. I will probably modify this function in the future, as I plan to work a bit more on this coefficients. Below are main reasons why I don't modify it now:
1. This function is not called by the user, the array it generates is already included in the repository. It can be used if you want to modify the array for some reason, but then you should know what you are doing.
2. Technically it should be valid independent from Z.
3. Z_min and Z_max are in fact defined in https://github.com/chalmersplasmatheory/DREAM/blob/master/include/DREAM/Equations/SlowingDownFrequency.hpp
In theory I can use e.g. cython to link this, but I am afraid that this will generate more problems than it solves.

PR #226, opened 2023-04-20
bug fix for when used with the linear implicit solver

peterhalldestam:
Problem identified!

The relevant equation term defined in `IonSpeciesTransientTerm.hpp`,  used when solving

$$
\frac{\text{d}W_i}{\text{d}t}=0,
$$

had a bug in its ´SetMatrixElement` (used with the linear implicit solver). It essentially solved the discretised equation

$$
\frac{W_i(t+\Delta t) - 1}{\Delta t} = 0,
$$

rather than the correct

$$
 \frac{W_i(t+\Delta t) - W_i(t)}{\Delta t} = 0.
$$

The former obviously yielded the solution $W_i=1$. This has now been corrected!

### DIFF
--- a/include/DREAM/Equations/Fluid/IonSpeciesTransientTerm.hpp
+++ b/include/DREAM/Equations/Fluid/IonSpeciesTransientTerm.hpp
@@ -3,7 +3,7 @@
 
 /**
  * Implementation of an equation term representing the
- * (backward euler) time derivative on an ion species 
+ * (backward euler) time derivative on an ion species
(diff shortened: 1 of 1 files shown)


hoppe93 (approved):
Wow, good catch! Not obvious how the non-linear solver managed to deal with that... :P

hoppe93:
Feel free to push the merge button (locating this bug proves you have sufficient experience with DREAM to gain merge-to-master rights ;)

PR #232, opened 2023-05-29
Redirect runiface stdout&stderr to files

20172931:


py/DREAM/DREAMTask.py
```diff
@@ -32,26 +32,57 @@ def __init__(self, settings, outfile=None, quiet=False, timeout=None, DREAMPATH=
         self.p = None
         self.obj = None
         self.stderr_data = None
+        #self.stderr_dat_byte=None
```
hoppe93, line 35: If this is no longer required it should probably be removed

py/DREAM/DREAMTask.py
```diff
@@ -32,26 +32,57 @@ def __init__(self, settings, outfile=None, quiet=False, timeout=None, DREAMPATH=
...
         self.DREAMPATH = DREAMPATH
         self.quiet = quiet
         self.timeout = timeout if timeout is not None else float('inf')
+        self.nthreads = nthreads
+        
+        self.stdout_name =stdout_name   
+        self.stderr_name =stderr_name  #new
```
hoppe93, line 42: What does "new" refer to here? (maybe remove?)

py/DREAM/DREAMTask.py
```diff
@@ -32,26 +32,57 @@ def __init__(self, settings, outfile=None, quiet=False, timeout=None, DREAMPATH=
...
+        self.open_stdout = None
+        self.open_stderr = None
 
     def run(self):
         self.startTime= time.time()
         if self.p != None: #if process is already created then we can safely od nothing
```
hoppe93, line 48: Typo: "od" -> "do"

py/DREAM/DREAMTask.py
```diff
@@ -32,26 +32,57 @@ def __init__(self, settings, outfile=None, quiet=False, timeout=None, DREAMPATH=
...
         try:
-            self.stderr_data = self.p.communicate(timeout=timeout)[1].decode('utf-8')
-
+            self.stderr_dat_byte=self.p.communicate(timeout=timeout)[1]
+            if self.stderr_dat_byte != None:
+                self.stderr_data = self.stderr_dat_byte.decode('utf-8')
+           
+            if self.stdout_name!='':
+                self.open_stdout.close()
+                print(f"one process finished")
```
hoppe93, line 79: We should probably make sure that ``self.quiet`` is ``True`` before printing. Also, I would prefer if this printing was done in ``runiface_parallel.py`` when the process is removed from the list of running tasks.

py/DREAM/DREAMTask.py
```diff
@@ -32,26 +32,57 @@ def __init__(self, settings, outfile=None, quiet=False, timeout=None, DREAMPATH=
...
+            if self.stderr_name!='':
+                self.open_stderr.close()
+                
             if self.p.returncode != 0:
                 self.errorOnExit = 1
-            else:
+            else:  #gives error when closing file since outfiles do not correspond to the actual output directory files
```
hoppe93, line 85: ??

py/DREAM/runiface.py
```diff
@@ -28,37 +27,44 @@ def locatedream():
...
     queue = []
     active = []
     allTasks = []
+    
     if len(settings) != len(outfiles):
         raise DREAMException("Lengths of settings and outfiles arrays are different!")
     
-    for _settings, outfile in zip(settings, outfiles):
-        task = DREAMTask(_settings, outfile, quiet, timeout, DREAMPATH )
+    #for _settings, outfile, stdout_name, stderr_name in zip(settings, outfiles,stdout_list,stderr_list):
```
hoppe93, line 62: This did not work when ``stdout_list`` and/or ``stderr_list`` were not given, so I had to replace it with the for loop below. You could remove this line now.

py/DREAM/runiface.py
```diff
@@ -28,37 +27,44 @@ def locatedream():
...
+    for i in range(len(settings)):
+        so = '' if len(stdout_list)==0 else stdout_list[i]
+        se = '' if len(stderr_list)==0 else stderr_list[i]
+        task = DREAMTask(settings[i], outfiles[i], quiet, timeout, DREAMPATH, so, se, nthreads=nthreads)
+           
         queue.append(task)
         allTasks.append(task)
```
hoppe93, line 70: Below here, in the if statement for ``task.hasFinished(0.1)``, you could do something like
```python
if not quiet:
    print(f" --> Process finished, output written to '{task.outfile}'.")
```

hoppe93 (changes requested):
Nice! I made some changes and added a bit of documentation (before going through your code). After going through the code I saw some potential improvements, which you maybe could fix?

hoppe93 (approved):

hoppe93 (approved):

PR #237, opened 2023-06-22
Partial screening modification

jjwalkowiak:
Added new model for partial screening
Separated partial screening from stopping power
Added runaway rate for Avalanche generation


src/Equations/PartialScreeningFactorWalkowiak.cpp
```diff
@@ -0,0 +1,72 @@
+#include<iostream>
```
hoppe93, line 1: I guess that these functions are described in a paper somewhere (perhaps https://doi.org/10.1063/5.0075859 ?) Maybe you could add a comment at the top of this which just briefly describes what is done and refers to the relevant paper?

hoppe93 (changes requested):
Looks good! If you could just add a comment at the top of ``PartialScreeningFactorWalkowiak.cpp`` which refers to a relevant paper, I think this should be ready to merge! Some of the automatic tests on GitHub utilize the partial screening functionality, suggesting that the old partial screening calculations work as expected.

src/Equations/PartialScreeningFactorWalkowiak.cpp
jjwalkowiak: Done

hoppe93 (approved):
Now all looks very good! :)

PR #238, opened 2023-06-28
Dreicer-nn-fix

liamanton:
Ni vet vad ni ska göra.


src/Equations/DreicerNeuralNetwork.cpp
```diff
@@ -82,13 +84,48 @@ real_t DreicerNeuralNetwork::RunawayRate(
...
+*exp(a * x^b)
+*/
+real_t DreicerNeuralNetwork::nn_fix(
+    const real_t EDD, const real_t logTheta, const real_t Zeff,
+    const real_t Zeff0, const real_t Z0Z, const real_t Z0_Z, const real_t logNfree, 
+    const real_t nfree_ntot, const real_t nfree, const real_t tauEE, bool use_fix
+) {
+    real_t boundary = 0.015;
+    real_t dE = 0.0001;
+    if(use_fix == true & EDD < boundary){
```
hoppe93, line 106: In C/C++ you should use double ampersands ``&&`` for logical AND in if statements. The single ampersand is a bit-wise AND operation, which in this case yields the same result.

src/Equations/DreicerNeuralNetwork.cpp
```diff
@@ -82,13 +84,48 @@ real_t DreicerNeuralNetwork::RunawayRate(
...
+        real_t gamma = 4.0/(3.0*M_SQRTPI) * RunawayRate_derived_params(
+            boundary, logTheta, Zeff, Zeff0, Z0Z, Z0_Z,
+            logNfree, nfree_ntot);
+        real_t gamma_dE = 4.0/(3.0*M_SQRTPI) * RunawayRate_derived_params(
+            boundary+dE, logTheta, Zeff, Zeff0, Z0Z, Z0_Z,
+            logNfree, nfree_ntot);
+
+        real_t A = log(gamma);
+        real_t B = (log(gamma_dE) - log(gamma))/dE;
```
hoppe93, line 115: I don't think this is really covered in numerics courses at Chalmers, but there are actually some neat tricks for evaluating numerical derivatives. The Numerical Recipes book (@IdaEkmark should have a copy) describes this in section 5.7. The conclusion though is basically that if you are constrained to using a forward difference, ``dE = sqrt(eps) * x_c``, where ``eps`` is the machine epsilon (~1e-16 in double precision floating-point) and ``x_c`` is the "characteristic scale" over which the function to differentiate varies. This can be difficult to estimate, so often one takes ``x_c = x``, which would be ``x_c = boundary`` in the notation used here.

If possible to use, central differences give better accuracy (but for the purposes here a forward difference is probably fine).

hoppe93 (changes requested):
Nicely done! I added some comments with tips and suggestions for improvements. In particular, I think it would be very nice to be able to toggle this functionality from the Python interface.

src/Equations/DreicerNeuralNetwork.cpp
IdaEkmark, line 1: (Probably not an important comment, but I'm curious) Now the derivative of gamma is not continuous, but the derivative of ln(gamma) is. Is there a large difference if you do 
B = (gamma_dE - gamma)/dE
b = B/A * boundary / gamma
This would solve
exp(a * boundary^b) = gamma
d(exp(a * EED^b)) / d EED (EED = boundary) = (gamma_dE - gamma) / dE

IdaEkmark (changes requested):
The first two comments should be addressed (and the third if you are curious ;) )

liamanton:
I think it is ready to be merged into the master branch now.

include/DREAM/Equations/RunawayFluid.hpp
```diff
@@ -160,6 +162,8 @@ namespace DREAM {
 
         void Rebuild(const real_t);
         void GridRebuilt();
+        bool GetextrapolateDreicer()
```
hoppe93, line 165: Could you change the name of this method to ``GetExtrapolateDreicer``, so that it is consistent with the naming of other methods in the code?

hoppe93 (changes requested):
Great! I just have some minor suggestions.

hoppe93 (approved):
Now I think this looks good! :)

Issue #246, opened 2023-08-10
Problem with extrapolation to last flux grid point in Svensson transport model


vaoskar:
It appears that the extrapolation of some quantities from the cell grid to the outermost point on the flux grid in SvenssonTransport.tcc becomes problematic if the gradient at the edge is too large. In particular, SPI simulations with large injected quantities seems to cause extrapolated values of the collision time to become negative. The extrapolated values should therefore be checked and, if necessary, corrected to ensure arithmetically and physically valid results.

PR #247, opened 2023-08-10
Fix issue #246

vaoskar:


vaoskar:
This pull request fixes issue #246 by taking a limiting value if the collision time extrapolates to zero on the last flux grid point (it might be relevant to have some check on Zeff as well, but hopefully this fixes the problem we are having at the moment)

hoppe93 (approved):
With the fix, this looks good now!

PR #248, opened 2023-08-10
Actually set w ( pitch-angle width factor) to zero if collision time …

vaoskar:
…extrapolates to negative values at the last flux grid point


hoppe93:
I thought we discussed this before and concluded that constant-extrapolation might be best? You probably have a reason to change this, and I'm curious to know why?

vaoskar:
I thought the conclusion was that we should keep the linear extrapolation
and only adjust the cases where it causes problems? That was what I tried
to do before as well, I only missed actually setting w=0, which enters into
more calculations than just the w_factor.

Den tors 10 aug. 2023 kl 16:05 skrev Mathias Hoppe ***@***.***

hoppe93 (approved):
Ah, ok, I see!

PR #249, opened 2023-09-04
Plasmoid drift

liamanton:


src/Equations/SPIHandler.cpp
```diff
@@ -324,6 +553,21 @@ void SPIHandler::Rebuild(real_t dt){
...
+                    shift_r[ip] = Deltar(ip);
+                    if(shift_r[ip] + xp[ip]>nr*Dr)
+                        shift_r[ip] = nr*Dr-xp[ip];
+                    nbrShiftGridCell[ip]=std::round(shift_r[ip]/Dr);
+                }
+            }
+        }
         // Shift the deposition profile
         for(len_t ip=0;ip<nShard;ip++){
             if(rCoordPNext[ip]>rCoordPPrevious[ip]){
```
vaoskar, line 573: I know this is mostly on me, but the loop condition here should be ir<nr-nbrShiftGridCell[ip] (before we have only used shifts of one grid cell and didn't plan well enough for the future...). Also, one should set depositionProfilesAllShards[ir*nShard+ip] = 0 for ir>nbrShiftGridCell[ip] (and make the corresponding changes for the rCoordPNext[ip]<rCoordPPrevious[ip]-case).

vaoskar (changes requested):
Really nice work! As far as I can tell this should work (except for some small mistakes that are mostly my fault...) for the test cases we have used! There are just a few simplifications of the implementation that would be good to alleviate to make the code work for a more general case before merging.

vaoskar (dismissed):
Great job! I only found some small things to change, and it felt easier to just fix them than writing detailed enough instructions for how to do it, so I fixed them myself. FYI I summarise the changes below:

There were a couple of places where you replaced Zavg0 where it shouldn't be replaced. 

I managed to make sure the drift is only calculated in the first iteration, so that we can simply get the conductivity directly from RunawayFluid in the first iteration and then just save it, which helps to make sure we use the same conductivity mode in both classes. 

I saw some errors in the logic on how to shift the deposition profiles, and some changes had to be done to ensure compatibility with the prescribed drift and the shift of the heat absorbed into the cloud.

Finally, you seemed to have made some potentially unintentional changes in the example running script, so I reverted it to use the same parameters as in the original case. I don't know what parameters you used in the example in the overleaf documentation, but perhaps you want to update that.

With these changes in place, it seems to work fine, as far as I can tell at least, so for me it is good to go! Does @hoppe93 want to take a look before we merge?

src/Equations/SPIHandler.cpp
```diff
@@ -541,6 +820,20 @@ void SPIHandler::CalculateIrp(){
     }
 }
 
+int SPIHandler::CalculateDriftIrp(len_t ip, real_t shift){
+    int temp = 0;
```
hoppe93, line 824: ``int_t``?

hoppe93 (changes requested):
Impressive work! :D I have a number of questions/comments though which will need to be addressed before we merge.

src/Equations/SPIHandler.cpp
```diff
@@ -221,13 +275,199 @@ void SPIHandler::DeallocateQuantities(){
...
+ * succeeding the pellet injection. The analytical model is derived in doi:10.1017/S0022377823000466
+ */
+
+ // Stores data about the surroundings of a shard
+void SPIHandler::AssignShardSpecificParameters(int ip){
+    v0 = -vp[3*ip];
+    n_e = ncoldPrevious[irp[ip]];
+    Te = TcoldPrevious[irp[ip]];
+    B = sqrt(this->rGrid->GetFSA_B2(irp[ip])) * rGrid->GetBmin(irp[ip]);
+    sigma = rf->GetElectricConductivity(irp[ip]);// As this function will only be called during the first newton iteration of each time step, sigma will be the conductivity from the previous time step
```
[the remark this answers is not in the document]
hoppe93: If ``RunawayFluid`` is rebuilt before ``SPIHandler`` you should get the conductivity for the current time step, and not for the "previous time step" as indicated in the comment. If you just want "some conductivity", this is fine, but the comment is inaccurate in that case.

(The ``RunawayFluid`` object is always rebuilt with the best available solution for the current time ``t``, so I would say it is "rebuilt based on values from the latest time step even if discarded". I don't see how that would be strange? To me that sounds like the expected behaviour of a Newton solver?)

src/Settings/SPIHandler.cpp
```diff
@@ -53,10 +67,28 @@ SPIHandler *SimulationGenerator::ConstructSPIHandler(FVM::Grid *g, FVM::UnknownQ
...
+    real_t *ZavgDriftArray = new real_t[nZavgDrift];
+    len_t *ZsDrift = new len_t[nZavgDrift];
+    len_t *isotopesDrift = new len_t[nZavgDrift];
+    if(spi_shift_mode == OptionConstants::EQTERM_SPI_SHIFT_MODE_ANALYTICAL){
+        for (len_t i = 0; i < nShard; i++)
+            TDrift[i] = (real_t)_TDrift[i];
+        for (len_t i = 0; i < nZavgDrift; i++){
+            ZavgDriftArray[i] = (real_t)_ZavgDriftArray[i];
+            ZsDrift[i] = (len_t)_ZsDrift[i];
+            isotopesDrift[i] = (len_t)_isotopesDrift[i];
```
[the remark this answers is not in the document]
hoppe93: (I will take care of this)

src/Settings/SPIHandler.cpp
```diff
@@ -53,10 +67,28 @@ SPIHandler *SimulationGenerator::ConstructSPIHandler(FVM::Grid *g, FVM::UnknownQ
         isotopes[i]=(len_t)_isotopes[i];
     }
     
-    len_t *nbrShiftGridCell = new len_t[nShard];
+    int_t *nbrShiftGridCell = new int_t[nShard];
     for (len_t i = 0; i < nShard; i++)
-        nbrShiftGridCell[i] = (len_t)_nbrShiftGridCell[i];
+        nbrShiftGridCell[i] = (int_t)_nbrShiftGridCell[i];
```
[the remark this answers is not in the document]
hoppe93: (I will take care of this)

hoppe93 (changes requested):
Some things still need addressing, in particular regarding memory handling and leaks.

src/Equations/SPIHandler.cpp
```diff
@@ -87,44 +109,74 @@ SPIHandler::SPIHandler(FVM::Grid *g, FVM::UnknownQuantityHandler *u, len_t *Z, l
...
-		 
-		        if(Z[iZ]==1 && isotopes[iZ]==2){
+                
+		        if(Z[iZ]==1 && isotopes[iZ]==2)
 		            pelletDeuteriumFraction[ip]+=molarFraction[offset+ip];
-		        }
+
+                ZavgDrift[ip]+=ZavgDriftList*molarFraction[offset+ip];
+                if (!((Z[iZ]==1 && isotopes[iZ]==2) || Z[iZ]==10) && counter2==0){
+                    printf("SPIHandler: Using other molarFractions other than deuterium and neon will lead to urealistic ablating.");
```
hoppe93, line 177: This needs to be turned into a warning (i.e. using the ``DREAM::IO::PrintWarning()`` function, with an appropriate warning ID).

hoppe93 (changes requested):
Found another issue which must be resolved.

src/Equations/SPIHandler.cpp
```diff
@@ -87,44 +109,74 @@ SPIHandler::SPIHandler(FVM::Grid *g, FVM::UnknownQuantityHandler *u, len_t *Z, l
...
     this->nr=rGrid->GetNr();
     this->nShard=unknowns->GetUnknown(id_Yp)->NumberOfMultiples();
+    
+    this->NZ=NZ;
 
     // Memory allocation
     AllocateQuantities();
     
+    // Ablation cloud quantities
+    this->TDrift=TDrift;
```
hoppe93, line 119: ``this->TDrift`` is allocated in ``AllocateQuantities()``, but then immediately overwritten here, leading to a memory leak. Is this really the intended behaviour? Should ``TDrift`` really be allocated in ``AllocateQuantities()``? (if not, the ``delete [] TDrift`` should probably be moved directly to the destructor?

src/Equations/SPIHandler.cpp
```diff
@@ -87,44 +109,74 @@ SPIHandler::SPIHandler(FVM::Grid *g, FVM::UnknownQuantityHandler *u, len_t *Z, l
...
+                }
+            }
             for(len_t ip=0;ip<nShard;ip++){
 		        pelletMolarMass[ip]+=molarMass*molarFraction[offset+ip];
 		        pelletMolarVolume[ip]+=molarMass/solidDensity*molarFraction[offset+ip];
-		 
-		        if(Z[iZ]==1 && isotopes[iZ]==2){
+                
+		        if(Z[iZ]==1 && isotopes[iZ]==2)
 		            pelletDeuteriumFraction[ip]+=molarFraction[offset+ip];
```
hoppe93, line 173: Why is this ``+=`` and not just ``=``? At this point, ``pelletDeuteriumFraction`` is uninitialized and could contain garbage values. It must therefore either be properly initialized first, or we should turn ``+=`` into ``=``.

src/Equations/SPIHandler.cpp
```diff
@@ -315,50 +583,112 @@ void SPIHandler::Rebuild(real_t dt){
...
-    }else if(spi_deposition_mode==OptionConstants::EQTERM_SPI_DEPOSITION_MODE_LOCAL_LAST_FLUX_TUBE){
-        CalculateTimeAveragedDeltaSourceLocal(depositionProfilesAllShards);
-
-        // Shift the deposition profile to the last grid cell (before the current one) to avoid "self-dilution"
+        // Shift the deposition profile
         for(len_t ip=0;ip<nShard;ip++){
-            if(rCoordPNext[ip]>rCoordPPrevious[ip]){
-                for(len_t ir=0;ir<nr-1;ir++){
-                    depositionProfilesAllShards[ir*nShard+ip]=rGrid->GetVpVol(ir+1)/rGrid->GetVpVol(ir)*depositionProfilesAllShards[(ir+1)*nShard+ip];
+            if(nbrShiftGridCell[ip]<0){
```
hoppe93, line 669: Somehow, the code is getting to this point without ``nbrShiftGridCell`` being initialized (it is allocated, but its elements have not been set to any particular values).

hoppe93 (changes requested):
Found some additional things which need addressing.

hoppe93 (approved):
This looks ready to merge now!

PR #252, opened 2023-10-02
Kinetic tritium Compton

IdaEkmark:
Implemented kinetic sources for the tritium and Compton runaway generation sources, as an alternative to the fluid rates already implemented.


hoppe93 (changes requested):
Looks mostly, good I just spotted some things!

src/Settings/RunawaySourceTerms.cpp
```diff
@@ -89,14 +91,54 @@ RunawaySourceTermHandler *SimulationGenerator::ConstructRunawaySourceTermHandler
...
+            real_t pLower;
+            if (hottailGrid != nullptr)
+                pLower = hottailGrid->GetMomentumGrid(0)->GetP1_f(hottailGrid->GetNp1(0));
+                
+            else if (runawayGrid != nullptr)
+                pLower = runawayGrid->GetMomentumGrid(0)->GetP1_f(0);
+            
+            if(grid == fluidGrid) {
+                oqty_terms->comptonSource_fluid = new ComptonSource(grid, unknowns, LoadDataT("eqsys/n_re/compton", s, "flux"), pLower, -1.0, ComptonSource::SOURCE_MODE_FLUID, REFluid);
+                rsth->AddSourceTerm(eqnSign + "fluid Compton", oqty_terms->comptonSource_fluid);
```
leferi99, line 105: I don't quite see a problem here but when I set COMPTON_RATE_ITER_DMS_KINETIC, the summary which comes up after the DREAM ASCII art says that I am using fluid Compton even though I have both hottail- and runaway grids. Maybe the problem is in src/Settings/Settings.cpp if that decides what will be printed to the terminal in the main.cpp?
@IdaEkmark

IdaEkmark: In your case, the quantity n_re should be interpreted as all the electrons with momentum above the lower limit of the runaway grid, and it is evaluated as a fluid based on the equation in the equation term information below the DREAM splash screen. However, importantly, this is just for keeping track of how many particles there are with higher momentum than the lower momentum limit - for the runaway electron distribution function f_re (from which the runaway current is also evaluated), it is the kinetic Compton source that is being used, not the fluid Compton source.

IdaEkmark: Great! Just let me know if you have any other questions :)

PR #253, opened 2023-10-04
Imas to master

olaszsoma:
The IDS interface for DREAM is implemented. It reads in the neccessary IDS fields and initializes a DREAM settings object, which can be used to setup a DREAM run. The output is not written into IDSs. Two separate options are available: either generate initial conditions based on the IDS provided, and run a self-consistent DREAM simulation, or follow the data prescribed in the provided IDS data and run a prescribed simulation. The implementation was tested on the Gateway system.
The DREAM documentation was also expanded with the information about the IMAS interface.


tools/imas/imas_wrapper.py
```diff
@@ -0,0 +1,542 @@
...
+		R = coreprof.vacuum_toroidal_field.r0
+		if log:
+			logging.info('The major radius of the magnetic axis from core profiles IDS loaded successfully!\n')
+
+	if len(coreprof.vacuum_toroidal_field.b0) == 0:
+		logging.warning('The vacuum toroidal field from core profiles IDS seems to be empty!\n')
+		raise Exception('Error in loading core profiles data. Please check the input data!')
+	else:
+		B0 = coreprof.vacuum_toroidal_field.b0
+		print(B0)
```
hoppe93, line 112: Maybe this ``print(B0)`` should be removed?

tools/imas/imas_wrapper.py
```diff
@@ -0,0 +1,542 @@
...
+	if len(coreprof.profiles_1d[0].grid.psi) == 0:
+		logging.warning('The poloidal flux from core profiles IDS seems to be empty!\n')
+		raise Exception('Error in loading core profiles data. Please check the input data!')
+	else:
+		cp_psi = coreprof.profiles_1d[0].grid.psi
+		if log:
+			logging.info('The poloidal flux from core profiles IDS loaded successfully!\n')
+	
+	# interpolate the coreprof rho from the coreprof psi to the equilibrium psi
+	interp_radius = CubicSpline(-cp_psi, cp_radius)
```
hoppe93, line 194: Does IMAS enforce a particular sign convention for poloidal flux? Or why do you negate ``psi`` here?

tools/imas/imas_wrapper.py
```diff
@@ -0,0 +1,542 @@
...
+			ds.eqsys.E_field.setType(ElectricField.TYPE_SELFCONSISTENT)
+			ds.eqsys.E_field.setBoundaryCondition(ElectricField.BC_TYPE_SELFCONSISTENT, inverse_wall_time=inverse_wall_time, R0=R)
+		
+		ds.eqsys.j_ohm.setInitialProfile(j_tor, radius=eq_radius, Ip0=Ip)
+		
+		return ds
+	
+	else:
+		print('return a list? hdf5?')
+		return 'TO DO'
```
hoppe93, line 249: This seems like something production code shouldn't do? It seems you have been setting a bunch of variables so far? If so, I would suggest return a ``dict`` with those variables here.

tools/imas/imas_wrapper.py
```diff
@@ -0,0 +1,542 @@
...
+
+import DREAM
+from DREAM.DREAMSettings import DREAMSettings
+import DREAM.Settings.Equations.IonSpecies as Ions
+import DREAM.Settings.Equations.ElectricField as ElectricField
+import DREAM.Settings.Equations.ColdElectronTemperature as Tcold
+from scipy.interpolate import CubicSpline
+
+
+def readInIDSSlice(shot, run, tokamak, user=os.getlogin(), time=-999, log=False, setUpDream=True, wall_radius=-999):
```
hoppe93, line 20: Could you add a comment describing what this function does? It is not immediately obvious how this differs from ``readInIDS()``.

tools/imas/imas_wrapper.py
```diff
@@ -0,0 +1,542 @@
...
+		return ds
+	
+	else:
+		print('return a list? hdf5?')
+		return 'TO DO'
+	
+	
+def readInIDS(shot, run, tokamak, user=os.getlogin(), log=False, setUpDream=True, wall_radius=-999):
+	
+	# TODO full IDS read in
```
hoppe93, line 254: Still TODO?

tools/imas/imas_wrapper.py
```diff
@@ -0,0 +1,542 @@
...
+	
+	# open shot file
+	dataEntry = imas.DBEntry(imasdef.MDSPLUS_BACKEND, tokamak, int(shot), int(run), user_name=user)
```
hoppe93, line 257: Which backends are available? If you have to specify the backend (which seems very weird to me, considering that the purpose of IMAS was to not use things like MDSplus) I guess it should be provided as an input parameter by the user?

tools/imas/imas_wrapper.py
```diff
@@ -0,0 +1,542 @@
...
+		ds.eqsys.T_cold.setType(Tcold.TYPE_PRESCRIBED)
+		
+		if e_field!=-999:
+			ds.eqsys.E_field.setType(ElectricField.TYPE_PRESCRIBED)
+			ds.eqsys.E_field.setPrescribedData(efield=cp_efield, radius=cp_radius[:,0], times=cp_time)
+		
+		else:
+			inverse_wall_time = resistivity / (4*np.pi*1e-7*R*np.log(R/wall_radius))
+			if log:
+				logging.info('The iverse wall time is %s 1/s with the given wall radius.\n', inverse_wall_time)
```
hoppe93, line 529: iverse -> inverse

tools/imas/imas_wrapper.py
```diff
@@ -0,0 +1,542 @@
...
+			ds.eqsys.E_field.setType(ElectricField.TYPE_SELFCONSISTENT)
+			ds.eqsys.E_field.setBoundaryCondition(ElectricField.BC_TYPE_SELFCONSISTENT, inverse_wall_time=inverse_wall_time, R0=R)
+		
+		ds.eqsys.j_ohm.setInitialProfile(eq_jtor[:,0], radius=eq_radius[:,0], Ip0=eq_ip[0])
+		
+		return ds
+	
+	else:
+		print('return a list? hdf5?')
+		return 'TO DO'
```
hoppe93, line 540: see my comment for the same location in ``readInIDSSlice()``

tools/imas/imas_wrapper.py
```diff
@@ -0,0 +1,542 @@
...
+	
+	dataEntry.close()
```
hoppe93, line 542: This will never be executed, right? You always return before this point?

tools/imas/imas_wrapper.py
```diff
@@ -0,0 +1,542 @@
...
+			ds.eqsys.E_field.setBoundaryCondition(ElectricField.BC_TYPE_SELFCONSISTENT, inverse_wall_time=inverse_wall_time, R0=R)
+		
+		ds.eqsys.j_ohm.setInitialProfile(j_tor, radius=eq_radius, Ip0=Ip)
+		
+		return ds
+	
+	else:
+		print('return a list? hdf5?')
+		return 'TO DO'
+
```
hoppe93, line 250: In ``readInIDS()`` you call ``dataEntry.close()`` at the end. Is that not needed here?

hoppe93 (changes requested):
Thanks for this Soma! Overall it looks good! I made some small modifications, primarily to the documentation, and I also have some comments and questions on the code.

I suppose an output interface would require an equal amount of work?

doc/sphinx/source/imas/index.rst
```diff
@@ -0,0 +1,64 @@
+
+DREAM interface for IMAS
+==========================
+In this section you will find the details of the DREAM-IMAS interface implementation. Use case leírás
```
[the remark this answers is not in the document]
olaszsoma: It was a "todo" I have accidentally left in. I removed it.

doc/sphinx/source/imas/index.rst
```diff
@@ -0,0 +1,67 @@
...
+
+The ``setUpDREAM()`` option can be used to initialize a DREAM settings object as an output from the function. The object initializes a DREAM run with the initial conditions for the physical quantities set from the IDS and the temperature and electric field calculated self-consistently. A few other settings might be required to be set independently from the IDS read in function. A simple example on how to use IDS's with DREAM can be found in the examples directory in the DREAM repository (``examples/imas``). This is a Python script which creates a ``dream_settings.h5`` file which can be used to run DREAM with the necessary input. The execution of the example file requires the ``imas`` Python module to be installed. The input generation from IDS's is conducted independently from running DREAM with the settings file, and can be run separately.
+
+The option not to create a DREAM settings object is currently not implemented, the loaded data could be saved in Python lists or dictionaries.
```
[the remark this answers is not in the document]
olaszsoma: Do you mean line 57? I am trying to say that currently it is not possible to run the interface without also generating a DREAMSettings object. It is probably not that useful as I assume most users would want to generate this anyway.
It also probably doesn't make sense to not create the settings object, as all the data is available in the input IDS if one knows where and how to get it. But it might be useful to add this if we think it useful.
Should I modify/remove this line?

tools/imas/imas_wrapper.py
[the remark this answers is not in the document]
olaszsoma: I think I added it in /tools after a discussion with you, but of course I can move it. 
I have done so, also modified the example files accordingly.

examples/imas/imas_test.py
```diff
@@ -0,0 +1,62 @@
...
+n0 = 4e18*np.ones(ds.radialgrid.nr)
+
+# rn, n0, rT, T0 = ...  get profiles of density and temperature
+ds.eqsys.f_hot.setInitialProfiles(rn0=r, n0=n0, rT0=rT, T0=ds.eqsys.T_cold.temperature)
+ds.eqsys.n_re.setHottail(Runaways.HOTTAIL_MODE_ANALYTIC_ALT_PC)
+
+# Disable runaway grid
+ds.runawaygrid.setEnabled(False)
+
+# Use the linear solver
```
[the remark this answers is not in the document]
olaszsoma: Comment modified.

examples/imas_prescribed/imas_test_prescribed.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env python3
+#
+# A basic example script on the imas wrapper for DREAM. This script generates a basic
+# DREAM input file which can be passed to 'dreami'. It is based on the basic.py example.
+#
+# Run as
+#
+#   $ ./imas_test.py
```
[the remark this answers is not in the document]
olaszsoma: Of course, fixed it.

examples/imas_prescribed/imas_test_prescribed.py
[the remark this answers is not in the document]
olaszsoma: Only the way the IDS is read in and the simulation results. In imas_test.py the IDS is used as an initial condition, i.e. one time slice is read in and the simulation is self-consistent after. I also add some argon to have a disruption in the imas_test.py example.
In the prescribed case the whole time evolution of the relevant IDS fields are read in and DREAM is run in a prescribed mode. In this case no runaways are generated if I remember correctly.

tools/imas/imas_wrapper.py
```diff
@@ -0,0 +1,542 @@
...
+from scipy.interpolate import CubicSpline
+
+
+def readInIDSSlice(shot, run, tokamak, user=os.getlogin(), time=-999, log=False, setUpDream=True, wall_radius=-999):
+	
+	# open shot file
+	dataEntry = imas.DBEntry(imasdef.MDSPLUS_BACKEND, tokamak, int(shot), int(run), user_name=user)
+	dataEntry.open()
+	
+	if log:
```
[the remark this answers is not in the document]
olaszsoma: I have implemented this suggestion. Now the log field of the function expects a string, which will be the name of the .log file. This information was added to the documentation.

tools/imas/imas_wrapper.py
olaszsoma: I was wondering why I got a number as an output from the example files. I removed it.

tools/imas/imas_wrapper.py
olaszsoma: I negate psi because the input psi function is a decreasing function and the interpolation requires a monotonically increasing one. I was looking how to solve this issue more elegantly, but this was what worked in the end.
Is there a better solution you can recommend?

doc/sphinx/source/imas/index.rst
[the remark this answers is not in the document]
olaszsoma: Based on one of your comments below, I have modified this line, as now the variables are saved in a dictionary.

tools/imas/imas_wrapper.py
olaszsoma: I agree and putting the into a `dict` seems the simplest solution. It is implemented, and the documentation is modified accordingly.

tools/imas/imas_wrapper.py
olaszsoma: This information is given in the documentation. If you prefer, I can add it here as well.

tools/imas/imas_wrapper.py
olaszsoma: Nope, forgot to remove. Done now.

tools/imas/imas_wrapper.py
olaszsoma: This is a bit outside my expertise, but if I recall correctly, the IDS structure can be defined in multiple backends, inluding MDSplus and HDF5 (I am not sure if there are any other). It is a remnant of old implementation of the CPOs, when they considered using other stuff than MDSplus, but they sticked with it. As far as I am aware only the MDSPLUS_BACKEND works with IMAS, and it has to be specified. The DBEntry function is not something I wrote, I am not sure this field can be circumvented.  We could ask Dmitriy Yadikin, if you want to do something about this.

tools/imas/imas_wrapper.py
olaszsoma: Thanks, fixed.

tools/imas/imas_wrapper.py
olaszsoma: Seen it, done it :).

tools/imas/imas_wrapper.py
olaszsoma: You are right, I added this to line 500 (in the newly edited file, see it after I uploaded everything), bfore the if-else part at the end of the function.

tools/imas/imas_wrapper.py
olaszsoma: No, it is already present, in line 47, after I read in the necessary time slices from the IDSs. This is done differently in the two functions, as the `readInIDS` function needs the whole time evolution of the various IDS fields, so I read in data throughout the loading of the different variables.
In `readInIDSSlice` just reads in one slice of the whole IDS field, and after doing so I extract the necessary variables from that field. I close the database after reading the slice in.
I hope this clears it up, if needed we can have a short discussion on this.

olaszsoma:
I have responded to you comments, and implemented changes where necessary.

tools/imas/imas_wrapper.py
hoppe93: Hmm... The sign of d(psi)/dr is typically quite important as it is chosen to follow a certain convention. Here it seems that you are interpolating from psi -> r, so the convention might not be too important, but shouldn't this just yield the wrong result? I think the desired solution is rather to sort the two arrays, i.e.
```
p = cp_psi.argsort()
cp_psi = cp_psi[p]
cp_radius = cp_radius[p]
```

tools/imas/imas_wrapper.py
hoppe93: I think it would be nice to have the info here as well, to avoid having to jump between the code and documentation.

tools/imas/imas_wrapper.py
hoppe93: Ok, if MDSplus is the only actual working backend then this is probably fine. Possibly we might want to add the backend as an option to ``readInIDSSlice()``, with this as default value? (I also noticed now that this is specified in at least one more place in this function)

tools/imas/imas_wrapper.py
olaszsoma: I am not interpolating from psi to r. I have a psi coordinate as a function of r in the Equilibrium IDS and the same in Core profiles. I get the midplane psi values from the 2D equilibrium, and interpolate the rho coordinates in Coreprof from the coreprof psi (which is not necessarily the midplane) to the equilibrium mid plane psi. The resulting rho is used as the radial coordinate in DREAM. Psi is later not used in the code, it is only used as an interpolation coordinate.

tools/imas/imas_wrapper.py
olaszsoma: Okay, I added the information to the code, for both fuctions.

tools/imas/imas_wrapper.py
olaszsoma: I didn't want to add it as an option to calling the functions as currently it does not take any other value as a valid option and I do not believe there are any plans to have another backend in the near future. I can add it if you need it, but I think it could be confusing, and would not have any purpose. It should be rather simple to add it if another backend is introduced.
Also, I am not sure what you meant by the sentence in the brackets. I did not find another place where it was specified in `readInIDSSlice()`, only in `readInIDS()`. Can you let me know in what line did you find it?

hoppe93: Ok, but then it sounds like maybe the backend is not needed as an input parameter for now.

(The lines where the MDSplus backend is referenced are 23 and 259)

tools/imas/imas_wrapper.py
hoppe93: Sorry, my point is that there are different conventions for what the sign of ``psi`` should be, and although for the particular dataset you have tested with now, it is negative and decreases towards the edge, that might not always be the case. If you just negate ``psi``, then it might work now for TCV, but maybe not for a different machine. So if the problem is that ``psi`` is expected to be monotonically increasing by the interpolation routine, then my suggestion would directly address this problem, regardless of the sign of ``psi``.

olaszsoma: I added this, as requested. I was not able to fully test it however, as the DREAM gave an error I was not able to fix. It was complaining about interpolation, once about x not monothonically increasing, once about x being out of range. I am not sure which variables were the problems, as it seemed a general error message from the interpolation method in c, without specifying where DREAM ran into problems. I will try to figure it out, but so far I was not successful.

hoppe93: That sounds like a GSL error, coming from the GSL interpolation routines. Basically, it means that some input data, which we interpolate in, has an x vector which is not monotonically increasing.

olaszsoma: Fixed them, everything works now.

hoppe93 (approved):
Thanks! I think this looks good now!

PR #257, opened 2023-10-23
Updated value of fine structure constant and C in tritium source

IdaEkmark:
When writing the Activated source derivation section of the paper, I realized the fine structure constant could have a higher precision, which also affected the normalization factor of the tritium beta decay energy distribution spectrum

### DIFF
--- a/src/Equations/Kinetic/TritiumSource.cpp
+++ b/src/Equations/Kinetic/TritiumSource.cpp
@@ -84,13 +84,13 @@ void TritiumSource::Rebuild(const real_t, const real_t, FVM::UnknownQuantityHand
  * ITER disruptions" by M. Solis (2017),
  */
 real_t TritiumSource::integrand(real_t p, void *){
-    real_t C = 1.2174246e-7; // Normalization factor
+    real_t C = 1.2174246e-7; // Normalization factor 
     real_t Tmax = 18.6e3;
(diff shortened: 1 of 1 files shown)


src/Equations/Kinetic/TritiumSource.cpp
```diff
@@ -84,13 +84,13 @@ void TritiumSource::Rebuild(const real_t, const real_t, FVM::UnknownQuantityHand
...
 real_t TritiumSource::integrand(real_t p, void *){
-    real_t C = 1.2176458e-7; // Normalization factor
+    real_t C = 1.2174246e-7; // Normalization factor
     real_t Tmax = 18.6e3;
     real_t mc2 = Constants::mc2inEV;
     if (p > sqrt((Tmax/mc2 + 1)*(Tmax/mc2 + 1) - 1)) {
         return 0;
     }
-    real_t alpha = 1./137.;
+    real_t alpha = 1./137.035999084;
```
hoppe93, line 93: For the fine structure constant, we actually already have it defined as
```
DREAM::Constants::alpha
```
in ``src/Constants.cpp``.

hoppe93 (approved):
Just added a note about the fine structure constant, otherwise looks fine!

hoppe93 (approved):
Super!

PR #259, opened 2023-10-23
Manually enforced lower physical limit on AMJUEL ionization loss

vaoskar:
This fix forces the combined ionization and line radiation losses given by the AMJUEL coefficients to be at least ionization energy times ionization rate (as it should not be physically possible to have any ionization events with a lower energy loss than the ionization energy). 

I will have to read the AMJUEL manual again tom make sure I have really understood the definitions of everything correctly. There might also be more accurate solutions to this, for instance the loss could be expressed in terms of the energy loss per effective ionization event (including intermediate excitations/deexcitations, thus accounting for line radiation), which AMJUEL has a separate, potentially more accurate, fit for. However, I think this is already an improvement at least, so I wanted to make this available and open the pull request to make it easier for you to review it.


hoppe93 (dismissed):
Looks good to me! @vaoskar if you wan to keep the pull request open until you've investigated further, feel free to do so. @IdaEkmark you could merge #257 to master and run your simulations with this branch instead (making sure to keep this branch up-to-date with master)

IdaEkmark (approved):
Looks good!

hoppe93 (dismissed):
Looks good! I'm happy (at least particularly so) if Ida can confirm that her simulations converge better with this.

hoppe93 (approved):
Looks good!

PR #261, opened 2023-11-03
Lcfs

lisehanebring:


include/DREAM/Equations/Fluid/LCFSLossRateTerm.hpp
```diff
@@ -0,0 +1,51 @@
...
+    private:
+    
+        FVM::UnknownQuantityHandler *unknowns;
+        real_t scaleFactor;
+        real_t* t_loss;
+        const len_t userGivenPsiEdge_t0;
+        real_t psi_edge_t0;
+        const len_t id_psi;
+        real_t* GammaLoss = nullptr;
+        len_t ir_LCFS;
```
vaoskar, line 21: len_t is unsigned, so if you want to set this variable to a negative value to indicate that there are no closed flux surfaces, you should change the type of this variable to int_t

src/Equations/Fluid/LCFSLossRateTerm.cpp
```diff
@@ -0,0 +1,166 @@
...
+    
+    // Find if the sign of PsiDiff() should be +1 or -1
+    if(this->signFixed == false) { 
+        real_t psi_mid_init = unknowns->GetUnknownInitialData(id_psi)[0]; 
+        real_t psi_edge_init = unknowns->GetUnknownInitialData(id_psi)[nr-1];
+        if(psi_mid_init-psi_edge_init > 0) { this->sign = 1; }
+        this->signFixed = true;
+    }
+    
+    bool exists = false;
```
vaoskar, line 82: A quite picky comment, but wouldn't it be possible to write lines 82-90 in a shorter way like

```
for(len_t ir = 0; ir < nr; ir++)
    if(this->sign * PsiDiff(ir)<0)
        this->ir_LCFS=(int_t)ir-1;
```
?

src/Equations/Fluid/LCFSLossRateTerm.cpp
```diff
@@ -0,0 +1,166 @@
...
+    return psi - this->psi_edge_t0;
+}
+
+
+
+/**
+* Currently a step function at ir_LCFS, could be changed to continuous
+*/
+real_t LCFSLossRateTerm::StepFunction(len_t ir){
+    real_t x;
```
vaoskar, line 109: A very picky comment but you don't need the variable x here, you can just write the content of this function as

```
    if (ir > this->ir_LCFS) { 
        return = 1; 
    } else { 
        return = 0; 
    }
```

vaoskar (changes requested):
Nice work! I only have a few very small and easy to fix-comments!

It did however occur to me that it might not have been intentional to put this implementation on the same branch as the hottail low-Z-implementation. Is the hottail low-Z-implementation ready to merge soon as well, or would it be worthwhile to remove it from this branch? I didn't review that part now as I'm not read up on how it works.

vaoskar (approved):
Looks good! You could just remove the outcommented lines and outdated comments I marked before merging. And then do whatever should be done with the hottail low-Z part and wait for Mathias approval.

src/Equations/Fluid/LCFSLossRateTerm.cpp
```diff
@@ -0,0 +1,208 @@
...
+            exists = true;
+        }
+    }
+    if ( exists == false ) { this->ir_LCFS = -1; } // No flux surface is closed
+}
+
+
+
+/**
+* Difference between the poloidal flux at radial point ir and at the edge at t=0
```
hoppe93, line 107: Doesn't this routine return the difference between ``psi_p`` at ``ir+1/2`` and the edge ``psi_p``?

src/Equations/Fluid/LCFSLossRateTerm.cpp
```diff
@@ -0,0 +1,208 @@
...
+*/
+real_t LCFSLossRateTerm::PsiDiff(len_t ir){
+    //real_t psi = unknowns->GetUnknownDataPrevious(id_psi)[ir]; 
+    // Interpolate (extrapolate for ir=0) psi for estimate of psi at inner 
+    // radial grid cell wall of current grid cell
+    real_t psi = InterpolatePsi(ir); 
+    return psi - this->psi_edge_t0;
```
hoppe93, line 114: I would suggest renaming ``psi`` -> ``psi_f`` to clarify that it is actually an edge value of the poloidal flux.

src/Equations/Fluid/LCFSLossRateTerm.cpp
```diff
@@ -0,0 +1,208 @@
...
+void LCFSLossRateTerm::SetWeights() {
+
+    SetGammaLoss(); // Set GammaLoss to be the weights on the radial grid
+    
+    len_t offset = 0;
+    for (len_t ir = 0; ir < nr; ir++){
+        const len_t xiIndex = this->GetXiIndexForEDirection(ir);
+        const real_t V = this->GetVolumeScaleFactor(ir);
+
+        weights[offset + n1[ir]*xiIndex] = scaleFactor * this->GammaLoss[ir] * V;
```
hoppe93, line 174: Are you sure this is how you want to implement it in kinetic mode? This will just remove particles from a single point in $(p,\xi_0)$. Wouldn't it be more natural to make it uniform across all of momentum space? In that case, the term should have the same value everywhere, and that value just has to be scaled appropriately with $\mathcal{V}'\Delta p\Delta\xi_0$.

py/DREAM/Output/LCFSLoss.py
```diff
@@ -0,0 +1,57 @@
...
+    
+        if r is None:
+            r=slice(None)
+            
+        if t is None:
+            t=slice(None)
+            
+        jreScrapedOff = self.output.eqsys.j_re.data[t,r]
+        
+        for i in range(len(self.grid.r)):
```
hoppe93, line 27: Shouldn't ``self.grid.r`` be replaced by the subset ``self.grid.r[r]``?

py/DREAM/Output/LCFSLoss.py
```diff
@@ -0,0 +1,57 @@
...
+            
+        if t is None:
+            t=slice(None)
+            
+        jreScrapedOff = CurrentDensity(name = self.name, data = self.calcScrapedOffJre(r,t), grid = self.grid, output=self.output)
+        
+        return jreScrapedOff.current(t = t)
+        
+    def plotScrapedOffJre(self, r=None, t=None, **kwargs):
+        jreScrapedOff = CurrentDensity(name = 'scraped-off $J_{re}$ [A/m$^2$]', data = self.calcScrapedOffJre(r,t), grid = self.grid, output=self.output)
```
hoppe93, line 52: Are you sure you want to give ``r`` and ``t`` to ``calcScrapedOffJre()``  and not to ``plot()``. I think you want the latter, as otherwise the radial/time grid will be screwed up (same comment for ``plotScrapedOffIre`` below).

src/Equations/SPIHandler.cpp
```diff
@@ -327,11 +327,11 @@ void SPIHandler::Rebuild(real_t dt){
         // Shift the deposition profile
         for(len_t ip=0;ip<nShard;ip++){
             if(rCoordPNext[ip]>rCoordPPrevious[ip]){
-                for(len_t ir=0;ir<nr-1;ir++){
+                for(len_t ir=0;ir<nr-nbrShiftGridCell[ip];ir++){
```
hoppe93, line 330: I suppose this is a borrow from the ``plasmoid_drift`` branch? But is it correct? Shouldn't ``ir`` be affected more directly? And do we really want these changes on this branch, before the ``plasmoid_drift`` branch is merged?

src/Settings/RunawaySourceTerms.cpp
```diff
@@ -152,7 +153,28 @@ RunawaySourceTermHandler *SimulationGenerator::ConstructRunawaySourceTermHandler
...
     }
+     
+    // Add LCFS loss term (add exception if (nr==1 and lcfs_mode != LCFS_LOSS_MODE_DISABLED)...)
+    OptionConstants::eqterm_lcfs_loss_mode lcfs_mode = (enum OptionConstants::eqterm_lcfs_loss_mode)s->GetInteger(mod + "/lcfs_loss");
+    if (lcfs_mode != OptionConstants::EQTERM_LCFS_LOSS_MODE_DISABLED && grid->GetNr() == 1){
+        throw SettingsException(
+            "The LCFS loss term is not compatible with using only one radial grid point."
+        );
+    }
+    len_t lcfs_user_input_psi = (len_t)s->GetInteger(mod + "/lcfs_user_input_psi");
```
hoppe93, line 169: Are you sure this shouldn't be a ``bool`` instead?

py/DREAM/Settings/Equations/RunawayElectrons.py
```diff
@@ -101,6 +118,51 @@ def setInitialProfile(self, density, radius=0):
...
+        
+    def setLCFSLossPsiEdget0(self, psi_edge_t0, user_input_active=1):
+        """
+        Sets the value of psi_p at the plasma edge at
+        t = 0, used to determine the LCFS radial point.
+        Use in restarts to keep the value from the first
+        simulation. user_input_active should be 1 (default)
+        or set to 0 to manually switch off the user input.
+        """
+        self.lcfs_user_input_psi = int(user_input_active)
```
hoppe93, line 151: Yeah, here it really looks like ``user_input_active`` should be a boolean instead!

hoppe93 (changes requested):
Nice work! I have some questions and suggestions for improvements here and there. Overall, I think this can be merged very soon! :)

py/DREAM/Settings/Equations/IonSpecies.py
```diff
@@ -72,7 +72,7 @@ def __init__(self, settings, name, Z, ttype=0, Z0=None, isotope=0, SPIMolarFract
         neutral_diffusion_mode=ION_NEUTRAL_DIFFUSION_MODE_NONE, neutral_prescribed_diffusion=None, rNeutralPrescribedDiffusion=None, tNeutralPrescribedDiffusion=None,
         charged_advection_mode=ION_CHARGED_ADVECTION_MODE_NONE, charged_prescribed_advection=None, rChargedPrescribedAdvection=None, tChargedPrescribedAdvection=None,
         neutral_advection_mode=ION_NEUTRAL_ADVECTION_MODE_NONE, neutral_prescribed_advection=None, rNeutralPrescribedAdvection=None, tNeutralPrescribedAdvection=None,
-        t_transp_expdecay_all_cs = None, t_transp_start_expdecay_all_cs = 0, diffusion_initial_all_cs = None, diffusion_final_all_cs = 0, advection_initial_all_cs = None, advection_final_all_cs = 0, r_expdecay_all_cs = None, t_expdecay_all_cs = None,        
+        t_transp_expdecay_all_cs = None, t_transp_start_expdecay_all_cs = 0, diffusion_initial_all_cs = None, diffusion_final_all_cs = 0, diffusion_offset_all_cs = 0, advection_initial_all_cs = None, advection_final_all_cs = 0, advection_offset_all_cs = 0, r_expdecay_all_cs = None, t_expdecay_all_cs = None,
```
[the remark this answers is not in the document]
vaoskar: It is supposed to be used to set a non-zero ion transport already before the onset of the later exponentially decaying transport coefficients. It was passed to the co-argument in calcTransportCoefficientExpdecayAllChargedStates() (and then forwarded to calcTransportCoefficientExpdecaySingleChargedState()) before, but I would think that disappeared when you restructured the setting of the ion transport and merged it to this branch from master?

py/DREAM/Settings/Equations/IonSpecies.py
```diff
@@ -1170,7 +1170,7 @@ def calcTransportCoefficientExpdecaySingleChargeState(self, t_exp, c0, cf = 0, t
         c_single_charge_state = (cf + np.exp(-(t-t_start)/t_exp)*(c0-cf))*(t>t_start)
         
         # Copy the coefficients for the last time step to avoid an unintended linear extrapolation with an unphysical sign change
-        c_single_charge_state = np.vstack((c_single_charge_state, c_single_charge_state[-1,:])) 
+        c_single_charge_state = co + np.vstack((c_single_charge_state, c_single_charge_state[-1,:]))
```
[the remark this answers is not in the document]
vaoskar: see above

py/DREAM/Settings/Equations/Ions.py
```diff
@@ -63,8 +63,8 @@ def addIon(self, name, Z, iontype=IONS_PRESCRIBED, Z0=None, isotope=0, SPIMolarF
         neutral_diffusion_mode=ION_NEUTRAL_DIFFUSION_MODE_NONE, neutral_prescribed_diffusion=None, rNeutralPrescribedDiffusion=None, tNeutralPrescribedDiffusion=None,
         charged_advection_mode=ION_CHARGED_ADVECTION_MODE_NONE, charged_prescribed_advection=None, rChargedPrescribedAdvection=None, tChargedPrescribedAdvection=None,
         neutral_advection_mode=ION_NEUTRAL_ADVECTION_MODE_NONE, neutral_prescribed_advection=None, rNeutralPrescribedAdvection=None, tNeutralPrescribedAdvection=None,
-        t_transp_expdecay_all_cs = None, t_transp_start_expdecay_all_cs = 0, diffusion_initial_all_cs = None, diffusion_final_all_cs = 0, advection_initial_all_cs = None, advection_final_all_cs = 0, r_expdecay_all_cs = None, t_expdecay_all_cs = None, 
-        init_equil=False, T=None, n=None, r=None, t=None, tritium=False, hydrogen=False):
+        t_transp_expdecay_all_cs = None, t_transp_start_expdecay_all_cs = 0, diffusion_initial_all_cs = None, diffusion_final_all_cs = 0, diffusion_offset_all_cs = 0, advection_initial_all_cs = None, advection_final_all_cs = 0, advection_offset_all_cs = 0,
```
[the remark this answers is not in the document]
vaoskar: see above

src/Equations/SPIHandler.cpp
lvotta98: Discuss!

py/DREAM/Settings/Equations/RunawayElectrons.py
```diff
@@ -91,6 +98,16 @@ def __init__(self, settings, density=0, radius=0, avalanche=AVALANCHE_MODE_NEGLE
         self.density = None
         self.radius  = None
         self.setInitialProfile(density=density, radius=radius)
+        
+        self.hottail_T_final = np.array([0])
+        self.hottail_T_final_r  = np.array([0])
```
hoppe93, line 103: These two lines are remnants from the hot-tail low-Z branch and should be deleted.

hoppe93 (changes requested):
Nice work! I found two more lines from the hot-tail low-Z branch which should be deleted. We should also fix all of Oskar's comments before merging.

hoppe93 (approved):
Great! This looks ready to merge now!

PR #268, opened 2024-01-18
Fix issue 265

peterhalldestam:
An optional parameter 'alpha' that can be set in the Python interface via the method Timestepper.setIonization has been added. Allowed values are in the closed interval [0,1] and defaults to 0 if not explicitly provided, and then the usual way of updating the timesteps will be used. If it takes a non-zero value, the  timestep will be set to alpha times the ionisation timescale (or to some 'dtmax' if this value exceeds this maximum value). This method seems to work well with my SPI simulations as far as I can see.

As of now, only this optional parameter 'alpha' is used to determine whether this way of updating the timesteps is to be used. Is this a good solution, or should an extra flag be added for more clarity, or any suggestions?


lvotta98 (approved):

hoppe93 (approved):
Looks excellent! And it's great that it seems to be working well in practice :D Feel free to merge whenever you feel ready!

While going through the code, I was wondering if maybe it could be useful to allow the user to store the estimated ionization time scale to the output file somehow? E.g. as an OtherQuantity? It can be reproduced with relative ease after the simulation by differentiating ``n_cold`` (although not necessarily with the same accuracy as during the simulation, if not all time steps are saved), so maybe a new class ``ColdElectronDensity`` in the Python output interface would suffice? Any opinions? (perhaps it's not really needed now)

hoppe93 (approved):
This should be ready to merge!

PR #271, opened 2024-02-18
Correct initialization of coronal equilibrium

hoppe93:
A bug was found in the implementation of coronal equilibrium initialization. This pull request fixes this bug.


hoppe93:
This pull request also:
- Adds notes for the calculation of a coronal equilibrium.
- Adds the kinetic ionization cross-section as an other quantity.
- Adds classes and routines for ion kinetic quantities in the Python output interface.
- Fixes a bug in the ``get_nist.py`` script which would always force a re-download of NIST data, if the data was already cached.

tools/get_nist.py
IdaEkmark, line 1: Can e be a unique name (ex 'D2' if there are two deuterium populations)? If not, this is fine. If yes, then should this be considered when checking that 'elements' is a subset of 'names'?

hoppe93: This script is not related to any actual simulations, but only loads a set of elements to read from ``tools/elements.json``. These are assumed to be unique (although no explicit check is performed :P) and must be valid, generally accepted, element names. That is, ADAS and NIST must agree on the name of the element.

PR #278, opened 2024-02-22
Memory leaks

IdaEkmark:
Especially when running many simulations with dreampyface, ~0.5GB of memory is lost per simulation, resulting in the python process being shut down when memory eventuelly is full. Have fixed some issues for dreampyface (still remain for runiface), but some remain


hoppe93 (approved):
After a few weeks of testing, these fixes seem to work well!

hoppe93 (approved):
Seems to be working fully now!

PR #279, opened 2024-02-22
Gamma spectrum

IdaEkmark:
Added funcionality of changing the gamma spectrum shape. The gamma spectrum will take the form 
log \Gamma_gamma = - exp(-z) - z + 1
z = [ln(W_\gamma/10^6)+C_1]/C_2 + C_3 * (W_\gamma/10^6)^2
(note that its the same as for ITER with C3=0)
where C1, C2 and C3 are parameters that can be specified in the setCompton python function. Default is ITER.


py/DREAM/Settings/Equations/RunawayElectrons.py
```diff
@@ -229,6 +254,12 @@ def fromdict(self, data):
                 self.comptonPhotonFlux  = data['compton']['flux']
                 self.comptonPhotonFlux_t = np.array([0.0])
 
+        if 'C1' in data['compton']:
+            self.C1_Compton = data['compton']['C1']
+            self.C2_Compton = data['compton']['C2']
+            self.C3_Compton = data['compton']['C3']
+            self.integratedComptonSpectrum = quad(GammafluxProfil, 0, np.inf, args=(self.C1_Compton, self.C2_Compton, self.C3_Compton))[0]
```
hoppe93, line 261: It would feel better if this was done in the C++ code (since anything could happen with this number between being calculated and actually used in the kernel), but I can see that this is more convenient for the programmer, and so maybe it's okay...

py/DREAM/Settings/Equations/RunawayElectrons.py
```diff
@@ -229,6 +254,12 @@ def fromdict(self, data):
                 self.comptonPhotonFlux  = data['compton']['flux']
                 self.comptonPhotonFlux_t = np.array([0.0])
 
+        if 'C1' in data['compton']:
+            self.C1_Compton = data['compton']['C1']
+            self.C2_Compton = data['compton']['C2']
+            self.C3_Compton = data['compton']['C3']
```
hoppe93, line 260: Since you're checking explicitly for ``float`` in the ``verify()`` method, you should explicitly typecast these inputs to floats (otherwise they may be e.g. ``numpy.float64``, which will not be tolerated by ``verify()``).

hoppe93 (changes requested):
Looks overall very good! I just have a few comments (only one of which really needs fixing), and probably we should add some documentation about how to use the new Compton parameters.

Issue #286, opened 2024-03-20
Type checking of C1, C2, C3 values fails for Compton gamma photon flux profile


leferi99:
`verifySettings()` in `DREAM/py/DREAM/Settings/Equations/RunawayElectrons.py` raises `EquationException("Invalid type for 'C1_Compton'. Expected float.")` error when restarting the simulation, due to the values being read from the HDF5 file.

leferi99:
Suggested solution is to change the `verifySettings()` function to check `type(self.C1_Compton) not in [float, np.double]` instead.

PR #287, opened 2024-03-21
Typecasting gamma photon flux coefficients in fromdict()

leferi99:
Implemented the suggestion from #286 to typecast gamma photon flux coefficients for the Compton source in `fromdict()` of `py/DREAM/Settings/Equations/RunawayElectrons.py` 

I tested it in my specific use-case and now it doesn't give the error anymore


hoppe93 (approved):
Great! Thanks for fixing! :D

PR #291, opened 2024-04-18
To get version used for Ekmark optimization paper 2024

IdaEkmark:
Should be no changes to the master branch, should only add the commit hash used for my optimization paper to the git history so that it can be used by others


hoppe93 (approved):
Looks perfect ;)

Issue #293, opened 2024-05-22
Pitch grid edges deallocated from memory if using `ds.hottailgrid.setCustomGrid(xi_f=...)` resulting in error in ther `dreami` executable


leferi99:
The error was
```
*** Error in `/DREAM/build/iface/dreami': double free or corruption (!prev): 0x0000000001de6f40 ***
```
The problem was in line 25 in "fvm/Grid/PXiGrid/XiCustomGridGenerator.cpp". Commenting this line and rebuilding DREAM fixed this issue.

hoppe93:
Also the ``PCustomGridGenerator`` class has the same bug.

PR #294, opened 2024-05-22
Fix memory error in custom grid generators

hoppe93:
This pull request fixes two memory errors reported in #293.


IdaEkmark (approved):
Looks great!

PR #297, opened 2024-05-24
Fix default dimension of SPI parameters

vaoskar:
Regards t_delay and nbrShiftGridCell


peterhalldestam:
The parameter `t_delay` is assumed by the kernel to be an array with one element for each shard. The user is expected to initialise this parameter as (1) `None`, (2) a scalar or (3) an array of length `rp.size`. To make this work, the following is done:

1. `None` -- which sets `t_delay` to be a zero array of size `rp.size`
2. A scalar value -- which is sets `t_delay` to be an array of size `rp.size` with every element set to this provided scalar value.
3. An array of size `rp.size` -- which is what we want!

peterhalldestam:
I added a check in `verifySettings` for the rare case where the user sets `t_delay` to an array with a different size than `rp.size`.

hoppe93 (approved):

PR #298, opened 2024-05-28
Approximate kinetic ionization mode

peterhalldestam:
Added ionization mode in which the runaway electron distribution is assumed to be mono-energetic delta-function. For more details, see doc/notes/kinioniz_approx.tex and examples/kinioniz_approx/compare.py for example usage.

The ionization mode is implemented as an additional mode to make it backwards compatible and can now be used both in fluid and kinetic simulations.


include/DREAM/OtherQuantityHandler.hpp
```diff
@@ -82,8 +83,10 @@ namespace DREAM {
 			// List of kinetic ionization rates for each ion species
 			std::vector<IonKineticIonizationTerm*> f_hot_kin_rates;
 			std::vector<IonKineticIonizationTerm*> f_re_kin_rates;
+            // List of approximated RE impact ionization rates for teach ion species
+            std::vector<IonFluidRunawayIonizationTerm*> f_re_kin_approx_rates;
```
hoppe93, line 87: Isn't it confusing if this variable has ``f_re`` in the name, even though it has nothing to do with ``f_re``? I would propose a name such as ``n_re_kin_rates`` instead.

py/DREAM/Settings/Equations/IonSpecies.py
```diff
@@ -43,6 +43,7 @@
 IONIZATION_MODE_FLUID = 1
 IONIZATION_MODE_KINETIC = 2
 IONIZATION_MODE_KINETIC_APPROX_JAC=3
+IONIZATION_MODE_APPROX_RE=4
```
hoppe93, line 46: I propose we name this ``IONIZATION_MODE_FLUID_RE`` instead.

src/Settings/Equations/ions.cpp
```diff
@@ -280,12 +281,14 @@ void SimulationGenerator::ConstructEquation_Ions(
     if (nZ0_prescribed > 0)
         ipp = new IonPrescribedParameter(fluidGrid, ih, nZ_prescribed, prescribed_indices, prescribed_densities);
 
+    bool includeKineticIonizationApprox = (ionization_mode == OptionConstants::EQTERM_IONIZATION_MODE_APPROX_RE);
```
hoppe93, line 284: I propose this variable is renamed ``includeFluidREIonization``.

src/Equations/Fluid/IonFluidRunawayIonizationTerm.cpp
```diff
@@ -0,0 +1,108 @@
...
+    AddUnknownForJacobian(u, id_ions);
+    AddUnknownForJacobian(u, id_nre);
+
+    preFactor = Constants::c * p0 / sqrt(1 + p0*p0);
+
+    nr = g->GetNr();
+    weights = new real_t[(Zion+1)*nr];
+
+    tableIndexIon = IonKineticIonizationTerm::GetTableIndex(Zion);
+    // possible to use an instance of IonKineticIonizationTerm, without having a MomentumGrid, to use the method for calculating sigma?
```
hoppe93, line 33: What does this comment mean?

hoppe93 (changes requested):
Very nice work! :D I only have some small comments in a few places, in particular on comments and the naming of the new mode.

Also, I noted that the vast majority of modifications were just additions of single spaces here and there. I suspect it's coming from some setting in your editor? While they don't cause any issues in the code at all, they make it a bit more difficult to navigate through the diffs on GitHub, so it would be nice if they could be avoided in future pull requests :)

hoppe93:
Also, could you merge in the latest master to this branch?

py/DREAM/Settings/OtherQuantities.py
```diff
@@ -35,6 +35,8 @@ class OtherQuantities:
         'fluid/Tcold_ion_coll',
         'fluid/W_hot',
         'fluid/W_re',
+        'fluid/kinioniz_approx_vsigma',
```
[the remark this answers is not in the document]
peterhalldestam: Good idea. I have changed it to `kinioniz_vsigma`, but this had the side effect of being identified together with `hottail/kinioniz_vsigma` and `runaway/kinioniz_vsigma` in `OtherQuantities.SPECIAL_TREATMENT` and tries to load it as another `OtherIonSpeciesKineticQuantity`. This raised an exception.

I figured the easiest solution was to add an if-statement in `OtherQuantities.setQuantity()` that checks if it's categorised as "fluid". I'm not sure if this is the best way to do it, so let me know if you have any suggestions! :)

hoppe93 (changes requested):
Very nice! Now the examples works (very well) for me. I just have one question on the ``OtherQuantities.py`` file, which should hopefully be very quick to sort out.

I also made a small (and unrelated to your changes) change of the ``tools/ADAS/rates.py`` file. It can simplify the use of those routines in various places, but does not impact any other areas of DREAM at the moment.

py/DREAM/Output/OtherQuantities.py
```diff
@@ -111,6 +111,8 @@ def setQuantity(self, name, data, attributes=None, datatype=None):
                 o = self.SPECIAL_TREATMENT[name](name=name, data=data, description=desc, grid=self.grid, output=self.output, momentumgrid=self.momentumgrid)
             elif data.ndim == 4 and self.momentumgrid is not None:
                 o = self.SPECIAL_TREATMENT[name](name=name, data=data, description=desc, grid=self.grid, output=self.output, momentumgrid=self.momentumgrid)
+            elif self.name == 'fluid':
```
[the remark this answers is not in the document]
peterhalldestam: It was in an effort to be able to keep the same name for the RE ionisation rate also used in the kinetic setups, but it was poorly implemented. An easier solution is to use a different name entirely, which I have done now!

hoppe93 (dismissed):
Looks good! :)

hoppe93 (approved):
Looks very good!

PR #299, opened 2024-05-28
Kinetic radiation other quantities

IdaEkmark:
Adds other quantities for Compton generation on RE-grid, synchrotron and bremsstrahlung radiation, as well as has some other bug fixes


py/DREAM/Output/DistributionFunction.py
```diff
@@ -195,17 +195,18 @@ def synchrotron(self, model='spectrum', B=3.1, wavelength=700e-9, t=None, r=None
 
         data = None
         if model == 'total':
+            data = np.zeros((len(t), len(r), self.data.shape[-2], self.data.shape[-1]))
```
hoppe93, line 198: Will ``len(t)`` and ``len(r)`` work of ``t`` or ``r`` is a slice? Maybe it would be safer to do ``self.grid.t[t]`` etc?

hoppe93 (changes requested):
Excellent! I just noticed one thing which may or may not need a fix.

hoppe93 (dismissed):
Excellent! I'm happy with this now :)

hoppe93 (approved):
Looks good and finally seems to work fully!

PR #303, opened 2024-06-01
Fix bug when not using the LCFS term

hoppe93:
A bug appeared which causes errors when loading settings from output files, and the LCFSLossRateTerm is not used in the simulation producing the output file. This commit fixes that bug.

It also fixes a number of smaller inconsistencies in the Python interface and Settings interface.


hoppe93:
Lorenzo will run a simulation to verify that these changes work.

lvotta98:
Need to merge master -> LCFS

PR #304, opened 2024-06-05
Add fitting routine for Compton spectrum

hoppe93:
Add fitting routine for Compton spectrum.


IdaEkmark:
Nice work! I have some comments:

- I think the flux G should be divided by bin size, such that the fitted spectrum integrates to the total flux (because right now the data sums to the total flux, which would mean that data of higher resolution would have lower values, right?)
- Should we have some constraint on the constants such that they are always positive? I think negative values would be bad, especially negative C_3 leads to the spectrum not tending to 0 with increasing energy
- Maybe add more descriptions about the dimensions expected of the data (most importantly that the energy should be in MeV, but also about the gamma flux data)
- Should there be an option to manually remove some data points that are too deviating? I had to do that for my data in order to get a good fit, although this can be done in the datafile as well which is easier...

hoppe93:
I made a number of changes based on the comments of @IdaEkmark:

- The input spectrum is now divided by the energy bin size by default. This behaviour can disabled via a command-line switch.
- The pre-factor and ``c_3`` coefficients are now constrained to non-negative values.
- A more extensive description was added to the ``--help`` flag.
- I added the flags ``--skip-below`` and ``--skip-above`` which allows the user to specify the indices for which to remove all values *below* or *above* (not inclusive). This seemed better than to specify all indices to remove individually, as in my case I had to remove the first ~16 points for a good fit.

IdaEkmark (approved):
Looks great!

Issue #306, opened 2024-06-11
MacOS Sonoma (14.5) unknown conversion error in building


TAWijkamp:
Building DREAM (master branch) on macOS Sonoma 14.5 (intel i9 processor macbook) raises the conversion error below. The same error is repeated on line 242 in ComptonSource.

ERROR log:

.../DREAM/src/Equations/Kinetic/ComptonSource.cpp:161:5: error: no matching function for call to 'gsl_integration_qng'
    gsl_integration_qng(&F, pm, pp, 0, 1e-8, &integral, &abserr, &neval);
    ^~~~~~~~~~~~~~~~~~~
/usr/local/Cellar/gsl/2.6/include/gsl/gsl_integration.h:176:5: note: candidate function not viable: no known conversion from 'len_t *' (aka 'unsigned long long *') to 'size_t *' (aka 'unsigned long *') for 8th argument
int gsl_integration_qng (const gsl_function * f,

TAWijkamp:
Works now, thanks!

PR #308, opened 2024-06-14
Fix compilation error

hoppe93:
Fix the compilation error on Mac OS X reported in issue #306.


IdaEkmark:
Nice work! I have some comments:
- I think the flux G should be divided by bin size, such that the fitted spectrum integrates to the total flux (because right now the data sums to the total flux, which would mean that data of higher resolution would have lower values, right?)
- Should we have some constraint on the constants such that they are always positive? I think negative values would be bad, especially negative C_3 leads to the spectrum not tending to 0 with increasing energy
- Maybe add more descriptions about the dimensions expected of the data (most importantly that the energy should be in MeV, but also about the gamma flux data)
- Should there be an option to manually remove some data points that are too deviating? I had to do that for my data in order to get a good fit, although this can be done in the datafile as well which is easier...

IdaEkmark:
Great!

IdaEkmark (approved):

Issue #309, opened 2024-06-14
Store warnings in output


hoppe93:
At the moment, warnings are easily missed as they are often printed to the start of the terminal output. A better way would perhaps be to also store the warnings in the output file and have the ``cli.py`` script print the warnings when the output file is opened.

PR #311, opened 2024-06-14
Save warnings to output file

hoppe93:
This pull request addresses issue #309.

Since warnings are often printed at the start of a simulation, they can easily be missed. This pull request implements functionality to save warnings to an output file, and have the interactive Python interface display those warnings when the output file is loaded.


IdaEkmark (commented):
One nice feature, that might be relevant if it is not to complicated to implement is to also print when the warning occurred (i.e. at which timestep). Is that a simple fix? Otherwise I think it looks good

PR #319, opened 2024-06-26
Fix bug in DREAM::IO

hoppe93:
A bug was introduced with the merge of the functionality to save warnings to the output file. The error was due to that, with the ``savewarnings`` branch, a new static variable ``DREAM::IO::simulation`` was introduced. When running multiple simulations in ``dreampyface``, this variable was not automatically reset at the start of each new run, and so it kept pointing to an old, deallocated simulation. When a warning was then emitted before ``DREAM::IO::Init()`` had been called (which is done at the end of ``SimulationGenerator::Process()``, which sets up all objects in the simulation), ``DREAM::IO::PrintWarning()`` would try to access the old deallocated ``Simulation`` object, leading to a segmentation fault.

The solution now introduced is to de-initialize ``DREAM::IO`` after solving the equation system, i.e. to set ``DREAM::IO::simulation = nullptr``. Since this object is only used for determining in which time step a warning is emitted, is only needed during the actually time stepping process and can then safely be set to a ``nullptr`` after time stepping has finished.

### DIFF
--- a/include/DREAM/IO.hpp
+++ b/include/DREAM/IO.hpp
@@ -31,6 +31,7 @@ namespace DREAM {
 		static Simulation *simulation;
 
 		static void Init(Simulation*);
+		static void Deinit();
 
         static bool VerifyMessage(const message_t);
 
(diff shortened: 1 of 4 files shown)


hoppe93:
This branch has also been tested on my previously failing ITER simulations, which now work well with this fix.

Issue #321, opened 2024-07-03
Coronal equilibrium problem for certain setups


peterhalldestam:
When running `examples/ionequil/generate.py` (master branch, debugging cmake mode) a "Floating-point error" is raised. This seems to originate in the method `SimulationGenerator::EvaluateIonEquilibrium`, where the ADAS coefficients are calculated with `nfree = 0` for some reason.

If I switch place on the two `addIon` methods in the the example script (so that the neon is added first, which is supposed to initialise in equilibrium), this error is not raised at all. Instead, the simulation output has both deuterium charge states set to zero.

However, when specifying both ion species to be initialised in coronal equilibrium, the simulation runs as expected.

peterhalldestam:
The problem seems to be due to some incorrect indexing in both in the equation initialiser in `src/Settings/Equations/ions.cpp` that lead to a miscalculation of `nfree`, and in `src/Settings/IonEquilibrium.cpp` where it essentially is assumed that the equilibrium species are always added first.

I pushed a possible fix to this to #322  and made a pull request! :D

PR #322, opened 2024-07-03
indexing fix

peterhalldestam:
possible fix to issue #321

### DIFF
--- a/src/Settings/Equations/ions.cpp
+++ b/src/Settings/Equations/ions.cpp
@@ -534,8 +534,9 @@ void SimulationGenerator::ConstructEquation_Ions(
 					ionOffset += Nr;
 				}
 			} else {
-				eq_ions.push_back(i);
+				eq_ions.push_back(dynamic_indices[i]);
 				vecni.push_back(ni+idx*Nr);
+				ionOffset += (Z+1)*Nr;
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


src/Settings/Equations/ions.cpp
```diff
@@ -549,7 +550,7 @@ void SimulationGenerator::ConstructEquation_Ions(
 			if (init_equil[iZ] != 0)
 				continue;
 
-			for (len_t Z0 = 1; Z0 <= ih->GetZ(iZ); Z0++, ionOffset++) {
+			for (len_t Z0 = 0; Z0 <= ih->GetZ(iZ); Z0++, ionOffset++) {
```
hoppe93, line 553: This modification (``1`` -> ``0``) is perhaps not necessary? For ``Z0=0``, the product on line 555 is always 0, so ``Z0=0`` will never provide a contribution?

hoppe93 (changes requested):
Very nice! Aside from one thing, I very much like your fix ;) (it's not exactly what I had in mind, but it's actually better)

src/Settings/Equations/ions.cpp
peterhalldestam: You're right. It was an attempt to update `ionOffset` correctly, but I realised I forgot to update it for the equilibrium ions. I believe I fixed this in commit 53d9839 (as well as avoiding the `Z0=0` case).

hoppe93 (approved):
Looks great now! :D

PR #323, opened 2024-07-05
Added interpolation for cylindrical plot of distribution function

IdaEkmark:
Added option for interpolation for cylindrical plot of distribution function, for p_perp=0


hoppe93 (approved):
Looks great!

PR #326, opened 2024-07-16
Integrate kinetic activated sources in fluid mode

IdaEkmark:
If you use the fluid activated sources without any kinetic grids, the kinetic sources are now integrated. We get the same as rates as with the already implemented fluid sources. 

There is a small difference for the tritium source between the fluid and integrated kinetic sources, but the integrated kinetic source agrees with the theory (when integrating the formula by Martin-Solis from other.fluid.pCrit), which suggests that it is the Taylor expansion used in the fluid source that gives the discrepancy. Interestingly, the theory agrees better with the fluid source when instead integrating from other.fluid.pStar, but I checked several times and this should be wrong.


hoppe93 (approved):
Looks very nice! :D

PR #329, opened 2024-07-31
Store equilibrium quantities in output file

hoppe93:
This branch implements functionality to store equilibrium quantities (primarily flux surface contours so far) in the output file. This can be used to generate more accurate visualizations of e.g. pellet shard trajectories.

### DIFF
--- a/fvm/Grid/CylindricalRadialGridGenerator.cpp
+++ b/fvm/Grid/CylindricalRadialGridGenerator.cpp
@@ -149,3 +149,99 @@ real_t CylindricalRadialGridGenerator::FindClosestApproach(
     return sqrt((x1+tc*(x2-x1))*(x1+tc*(x2-x1))+(y1+tc*(y2-y1))*(y1+tc*(y2-y1)));
 }
 
+
+/**
+ * Returns a list flux surface R coordinates
+ * on the simulation grid.
+ */
+const real_t *CylindricalRadialGridGenerator::GetFluxSurfaceROverR0() {
+	const len_t nr = this->GetNPsi();
+	const len_t ntheta = this->GetNTheta();
+	real_t *R = new real_t[nr*ntheta];
+
+	for (len_t j = 0, i = 0; j < ntheta; j++) {
+		real_t theta = 2*M_PI*j / ntheta;
+
+		for (len_t ir = 0; ir < nr; ir++, i++)
+			R[i] = 1 + this->x[ir] * cos(theta);
+	}
+
+	return R;
+}
+
+
+/**
+ * Returns a list flux surface R coordinates
+ * on the simulation grid.
+ */
+const real_t *CylindricalRadialGridGenerator::GetFluxSurfaceROverR0_f() {
+	const len_t nr = this->GetNPsi();
+	const len_t ntheta = this->GetNTheta();
+	real_t *R = new real_t[(nr+1)*ntheta];
+
+	for (len_t j = 0, i = 0; j < ntheta; j++) {
+		real_t theta = 2*M_PI*j / ntheta;
+
+		for (len_t ir = 0; ir < nr+1; ir++, i++)
+			R[i] = 1 + this->x_f[ir] * cos(theta);
+	}
+
+	return R;
+}
+
+
+/**
+ * Returns a list flux surface Z coordinates
+ * on the simulation grid.
+ */
+const real_t *CylindricalRadialGridGenerator::GetFluxSurfaceZ() {
+	const len_t nr = this->GetNPsi();
+	const len_t ntheta = this->GetNTheta();
+	real_t *Z = new real_t[nr*ntheta];
+
+	for (len_t j = 0, i = 0; j < ntheta; j++) {
+		real_t theta = 2*M_PI*j / ntheta;
+
+		for (len_t ir = 0; ir < nr; ir++, i++)
+			Z[i] = this->x[ir] * sin(theta);
+	}
+
+	return Z;
+}
+
+
+/**
+ * Returns a list flux surface Z coordinates
+ * on the simulation grid.
+ */
+const real_t *CylindricalRadialGridGenerator::GetFluxSurfaceZ_f() {
+	const len_t nr = this->GetNPsi();
+	const len_t ntheta = this->GetNTheta();
+	real_t *Z = new real_t[(nr+1)*ntheta];
+
+	for (len_t j = 0, i = 0; j < ntheta; j++) {
+		real_t theta = 2*M_PI*j / ntheta;
+
+		for (len_t ir = 0; ir < nr+1; ir++, i++)
+			Z[i] = this->x_f[ir] * sin(theta);
+	}
+
+	return Z;
+}
+
+
+/**
+ * Returns the poloidal angle array corresponding
+ * to the 'GetFluxSurface()' methods.
+ */
+const real_t *CylindricalRadialGridGenerator::GetPoloidalAngle() {
+	const len_t ntheta = this->GetNTheta();
+	real_t *theta = new real_t[ntheta];
+
+	for (len_t i = 0; i < ntheta; i++)
+		theta[i] = 2*M_PI*i / ntheta;
+	
+	return theta;
+}
+
+
--- a/include/FVM/Grid/AnalyticBRadialGridGenerator.hpp
+++ b/include/FVM/Grid/AnalyticBRadialGridGenerator.hpp
@@ -97,6 +97,14 @@ namespace DREAM::FVM {
 		virtual void GetRThetaPhiFromCartesian(real_t*, real_t*, real_t*, real_t, real_t, real_t, real_t, real_t ) override;
 		virtual void GetGradRCartesian(real_t*, real_t , real_t, real_t ) override;
 
+		virtual const real_t GetZ0() override { return 0; }
+		virtual const len_t GetNPsi() override { return this->GetNr(); }
+		virtual const len_t GetNTheta() override { return 120; }
+		virtual const real_t *GetFluxSurfaceROverR0() override;
+		virtual const real_t *GetFluxSurfaceROverR0_f() override;
+		virtual const real_t *GetFluxSurfaceZ() override;
+		virtual const real_t *GetFluxSurfaceZ_f() override;
+		virtual const real_t *GetPoloidalAngle() override;
     };
 }
 
--- a/include/FVM/Grid/CylindricalRadialGridGenerator.hpp
+++ b/include/FVM/Grid/CylindricalRadialGridGenerator.hpp
@@ -45,6 +45,15 @@ namespace DREAM::FVM {
             {Jacobian=x[ir]; B=B0; NablaR2 = 1; ROverR0 = 1;}
         virtual void EvaluateGeometricQuantities_fr(const len_t ir, const real_t, real_t &B, real_t &Jacobian, real_t &ROverR0, real_t &NablaR2) override
             {Jacobian=x_f[ir]; B=B0; NablaR2 = 1; ROverR0 = 1;}
+
+		virtual const real_t GetZ0() override { return 0; }
+		virtual const len_t GetNPsi() override { return this->GetNr(); }
+		virtual const len_t GetNTheta() override { return 120; }
+		virtual const real_t *GetFluxSurfaceROverR0() override;
+		virtual const real_t *GetFluxSurfaceROverR0_f() override;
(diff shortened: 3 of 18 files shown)


hoppe93:
This needs to be tested with the ``ANALYTIC`` and ``CYLINDRICAL`` radial grids before merging!

py/DREAM/Output/SPIShardPositions.py
```diff
@@ -15,6 +17,70 @@ def __init__(self, name, data, grid, output, attr=list()):
...
+        """
+        Estimate the time at which the pellet arrives to the plasma edge.
+        """
+        if 'eq' not in self.grid:
+            raise OutputException("Cannot plot poloidal trajectory when equilibrium data is not stored in output.")
+
+        R0 = self.grid.eq.R0
+        if np.isinf(R0): R0 = 1
+
+        ROverR0 = self.grid.eq.ROverR0_f*R0 - R0
```
IdaEkmark, line 30: Is `ROverR0` an accurate name here? Is maybe `RminusR0` better? Or simply `R`, to be consistent with `Z` below? Although, `Z` could be renamed `ZminusZ0` as well

py/DREAM/Output/SPIShardPositions.py
```diff
@@ -15,6 +17,70 @@ def __init__(self, name, data, grid, output, attr=list()):
...
+
+        vertices = [(ROverR0[i,-1], Z[i,-1]) for i in range(ntheta)]
+        p = path.Path(vertices)
+
+        xp = self.data[:,0::3,0]
+        yp = self.data[:,1::3,0]
+
+        # Check if the shards start within the plasma
+        if p.contains_point((xp[0,0], yp[0,0])):
+            return 0
```
IdaEkmark, line 42: I assume that it is correct to just check if the first shard is in the plasma, and if it is not, no other shard will be either? Do they always start in the same position initially?

py/DREAM/Output/SPIShardPositions.py
```diff
@@ -15,6 +17,70 @@ def __init__(self, name, data, grid, output, attr=list()):
...
+
+        # Find the pellet which is travelling the fastest in the x direction
+        imax = np.argmax(np.abs(xp[-1,:] - xp[0,:]))
```
IdaEkmark, line 45: Do all pellets have the same speed at all times? Otherwise, why not use `xp[1,:]` instead of `xp[-1,:]`?

py/DREAM/Output/SPIShardPositions.py
```diff
@@ -59,6 +125,105 @@ def calcRadialCoordinate(self, shards=None, t=None):
...
+    def plotAtTime(self, t=-1, shards=None, ax=None, show=None):
+        """
+        Plot the pellet shards over the poloidal cross-section at a given time.
+        """
+        black = (87/255, 117/255, 144/255)
+        red = (249/255, 65/255, 68/255)
+
+        genax = ax is None
+
+        if genax:
```
IdaEkmark, line 139: Same comment about `genax` as above, i.e. is it necessary to define a separate variable?

py/DREAM/Output/SPIShardPositions.py
```diff
@@ -59,6 +125,105 @@ def calcRadialCoordinate(self, shards=None, t=None):
...
+            raise OutputException("Cannot plot poloidal trajectory when equilibrium data is not stored in output.")
+
+        eq = self.grid.eq
+
+        xp = self.data[:,0::3,0]
+        yp = self.data[:,1::3,0]
+
+        eq.visualize(ax=ax, shifted=True, maxis=False)
+
+        ax.plot(xp[0,0], yp[0,0], 'o', color=red)
```
IdaEkmark, line 158: Once again, do the shards always start in the same position initially?

py/DREAM/Output/SPIShardPositions.py
```diff
@@ -59,6 +125,105 @@ def calcRadialCoordinate(self, shards=None, t=None):
...
+        ax.plot(xp[t,shards], yp[t,shards], 'k.')
+
+        if np.isinf(eq.R0):
+            ax.set_xlabel('Radius $R-R_0$ (m)')
+        else:
+            ax.set_xlabel('Major radius $R$ (m)')
```
IdaEkmark, line 164: With `shifted=True` in `eq.visualize`, the x-axis will be $R-R_0$ regardless of what $R_0$ is, right?

py/DREAM/Output/SPIShardPositions.py
```diff
@@ -59,6 +125,105 @@ def calcRadialCoordinate(self, shards=None, t=None):
...
+    def plotTrajectoryPoloidal(self, shards=None, ax=None, show=None, color=None):
+        """
+        Plot the trajectory of one or more shards in a poloidal cross-section.
+        """
+        black = (87/255, 117/255, 144/255)
+        red = (249/255, 65/255, 68/255)
+
+        genax = ax is None
+
+        if genax:
```
IdaEkmark, line 184: Same comment about `genax` as above, i.e. is it necessary to define a separate variable?

py/DREAM/Output/SPIShardPositions.py
```diff
@@ -59,6 +125,105 @@ def calcRadialCoordinate(self, shards=None, t=None):
...
+
+        if shards is None:
+            i1 = np.argmin(yp[-1,:])
+            i2 = np.argmax(yp[-1,:])
+            shards = (i1, i2)
+
+            if color is None:
+                color = red
+
+        #ax.plot(R0*eq.ROverR0_f[:,-1] - R0, eq.Z_f[:,-1] - eq.Z0, color=black, linewidth=2)
```
IdaEkmark, line 206: Can we remove this comment?

py/DREAM/Output/SPIShardPositions.py
```diff
@@ -59,6 +125,105 @@ def calcRadialCoordinate(self, shards=None, t=None):
...
+        ax.plot(xp[0,0], yp[0,0], 'o', color=red)
+        if color is None:
+            ax.plot(xp[:,shards], yp[:,shards])
+        else:
+            ax.plot(xp[:,shards], yp[:,shards], color=color)
+
+        if np.isinf(eq.R0):
+            ax.set_xlabel('Radius $R-R_0$ (m)')
+        else:
+            ax.set_xlabel('Major radius $R$ (m)')
```
IdaEkmark, line 218: As above, with `shifted=True` in `eq.visualize`, the x-axis will be $R-R_0$ regardless of what $R_0$ is, right?

src/DiagonalPreconditioner.cpp
IdaEkmark, line 1: Indentation?

IdaEkmark (changes requested):
I have a couple of questions - I think not all of them will require changes, but some might

fvm/Grid/CylindricalRadialGridGenerator.cpp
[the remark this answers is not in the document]
hoppe93: Very good spotting ;) I think I intentionally made this a bit inconsistent. As you correctly point out, this all becomes weird in cylindrical geometry since $R_0\to\infty$. The goal was to still be able to give some R coordinates for the cylindrical geometry, which is technically not possible.

Perhaps a more consistent approach would be to give $(R-R_0)/R_0$ instead (this is what's done in for example LUKE)? Then the flux surfaces can be represented consistently in both toroidal and cylindrical geometry. Then this quantity should probably renamed though?

include/FVM/Grid/AnalyticBRadialGridGenerator.hpp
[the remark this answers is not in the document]
hoppe93: Currently there is no possibility to choose it, indeed, but there is nothing fundamental limiting this. In principle we could allow this, but for now, for simplicity, I did not since I didn't foresee that anyoone would need to change it. But maybe it's better to do it general from the start?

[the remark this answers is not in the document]
IdaEkmark: To at least have a parameter for it, whose value is hard coded, might be better just in case the need arises in the future, but I'm fine with how it is now as well.

fvm/Grid/CylindricalRadialGridGenerator.cpp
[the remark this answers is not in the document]
IdaEkmark: I don't quite see how this solves the problem though, as with $R=R_0+r$ ($r$ being the minor radius-vector projected in the horizontal plane), then $(R-R_0)/R_0=r/R_0=0$ if $R_0=\infty$. Do you mean just $R-R_0$?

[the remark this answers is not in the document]
hoppe93: Yes, that would be better! I'll try to change it!

include/FVM/Grid/CylindricalRadialGridGenerator.hpp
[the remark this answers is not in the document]
hoppe93: In this file I don't see that the identation is off (but maybe I changed my GitHub settings to use 4 spaces-per-tab?). Normally I let this pass, but maybe we should be stricter?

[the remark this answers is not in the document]
hoppe93: (in this case it's probably due to a mixture of tabs and spaces)

py/DREAM/Output/Equilibrium.py
```diff
@@ -0,0 +1,79 @@
...
+        :param shifted: If ``True``, shifts the surfaces so that (0, 0) is on the magnetic axis.
+        """
+        red  = (249/255, 65/255, 68/255)
+        black = (87/255, 117/255, 144/255)
+        gray = (190/255, 190/255, 190/255)
+
+        # Set up axes (if not already done)
+        genax = ax is None
+
+        if genax:
```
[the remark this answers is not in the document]
hoppe93: We use this in a lot of places in the output interface, and at one point in time it had a purpose, but now I don't there's any reason for this variable anymore :P

fvm/Grid/CylindricalRadialGridGenerator.cpp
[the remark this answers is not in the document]
hoppe93: (For Z we don't need any normalization, but for consistency it is perhaps better to define $Z-Z_0$ as well)

py/DREAM/Output/SPIShardPositions.py
hoppe93: All shards are launched from the same origin, so the idea here is that if one shard is inside the plasma, all are (because they are in the same place). It wouldn't work for a continuation simulation though, where the shards have already traveled a bit, so maybe we should generalize this as well?

py/DREAM/Output/SPIShardPositions.py
hoppe93: In the first simulation, yes. In contrast to your previous comment though, here I don't see any easy way of calculating the actual origin (because the pellet shatter-point is (sadly) not stored in the settings file).

IdaEkmark: Could we maybe have an if-statement, such that if the shards don't start in the same location, we can just print a warning about there being no information about the origin and skip the plotting of the origin, instead of assuming it starts in the first shards initial position?

py/DREAM/Output/Equilibrium.py
```diff
@@ -40,29 +40,25 @@ def visualize(self, shifted=False, maxis=True, ax=None, show=None, **kwargs):
...
         # Limiter
-        ax.plot(R0*self.ROverR0_f[:,-1]-rn, self.Z_f[:,-1]-zn, color=black, linewidth=2, **kwargs)
+        ax.plot(self.RMinusR0_f[:,-1]+rn, self.ZMinusZ0_f[:,-1]+zn, color=black, linewidth=2, **kwargs)
         # Magnetic axis 
         if maxis:
-            ax.plot(R0-rn, self.Z0-zn, 'x', color=red)
+            ax.plot(rn, zn, 'x', color=red)
         ax.axis('equal')
 
         if np.isinf(self.R0):
```
IdaEkmark, line 64: `if np.isinf(self.R0) or shifted:`?

IdaEkmark (commented):
Just a small comment

hoppe93:
> This needs to be tested with the `ANALYTIC` and `CYLINDRICAL` radial grids before merging!

This has now been checked and confirmed to work. I also improved the plotting by ensuring that the flux surfaces are closed.

IdaEkmark (approved):
Great!

PR #332, opened 2024-08-20
Fixed pyface save settings bug

IdaEkmark:
Added to check if data-object is a function when saving settings file, and then saving the key of the data-object as 1 to indicate that such a function is used, primarily for using terminate-functions


hoppe93 (approved):
Looks good!

PR #333, opened 2024-08-22
Added some features that I want

IdaEkmark:
Added a semilogx plot and a plotting routine to see the data with indices for each data point before fitting. Also moved the normalization of the flux to bin size before removing indices, and removed a print statement.


lvotta98 (approved):
Looks nice to me!

PR #336, opened 2024-09-20
Bug in DisruptionsActivated

IdaEkmark:
Fixed a bug in DisruptionsActivated-example, and also in KineticQuantity-plotting routine


hoppe93 (approved):
Great!

PR #346, opened 2024-10-18
Iss296

IdaEkmark:
I added a time object to the solver, so that the solver-plot-functions don't break when not saving data in all time-steps.

### DIFF
--- a/py/DREAM/Output/SolverNonLinear.py
+++ b/py/DREAM/Output/SolverNonLinear.py
@@ -69,7 +69,7 @@ class SolverNonLinear(Solver):
         This method returns an array of tuples, where each tuple denotes a
         single range of time steps where the backup solver was used.
         """
-        r = np.linspace(1, self.output.grid.t.size, self.output.grid.t.size)[np.where(self.backupinverter)]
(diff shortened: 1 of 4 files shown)


py/DREAM/Output/SolverNonLinear.py
```diff
@@ -17,6 +17,8 @@ def __init__(self, solverdata=None, output=None):
         """
         super().__init__(solverdata, output)
 
+        if 'solvertime' in solverdata:
+            self.solvertime = np.array([float(x) for x in solverdata['solvertime'][:]])
```
hoppe93, line 21: You shouldn't have to do this. It should possible to just do ``self.solvertime = solverdata['solvertime'][:]`` since the values should already be ``float``.

hoppe93: (Probably there's also a much better way of doing this for ``self.iterations``, by having ``numpy`` do the type conversion... I feel stupid now)

hoppe93 (changes requested):
This is really great! Just have one suggestion.

hoppe93 (dismissed):
Great!

hoppe93 (approved):
Greater!

Issue #347, opened 2024-10-24
Missing input parameter in other quantities


hoppe93:
Some other quantity constructors in the Python output interface are missing the input parameter ``momentumgrid``, which is expected by certain other quantities.

PR #348, opened 2024-10-24
Fix issue #347

hoppe93:
Add a missing input parameter to the constructors of certain other quantities in the output interface.

### DIFF
--- a/py/DREAM/Output/LCFSLoss.py
+++ b/py/DREAM/Output/LCFSLoss.py
@@ -8,7 +8,7 @@ from .ScalarQuantity import ScalarQuantity
 
 class LCFSLoss(OtherFluidQuantity):
 
-    def __init__(self, name, data, description, grid, output):
+    def __init__(self, name, data, description, grid, output, momentumgrid):
         """
         Constructor.
         """
@@ -54,4 +54,4 @@ class LCFSLoss(OtherFluidQuantity):
         
     def plotScrapedOffIre(self, r=None, t=None, **kwargs):
         jreScrapedOff = ScalarQuantity(name = 'scraped-off $I_{re}$ [A]', data = self.calcScrapedOffIre(), grid = self.grid, output=self.output)
-        return jreScrapedOff.plot(r=r, t=t, **kwargs)
+        return jreScrapedOff.plot(t=t, **kwargs)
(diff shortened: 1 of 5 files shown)


py/DREAM/Output/LCFSLoss.py
```diff
@@ -54,4 +54,4 @@ def plotScrapedOffJre(self, r=None, t=None, **kwargs):
         
     def plotScrapedOffIre(self, r=None, t=None, **kwargs):
```
peterhalldestam, line 55: The argument ``r`` is not used anymore, so it could be removed from this function's arguments

py/DREAM/Output/OtherQuantities.py
```diff
@@ -112,7 +112,7 @@ def setQuantity(self, name, data, attributes=None, datatype=None):
             elif data.ndim == 4 and self.momentumgrid is not None:
                 o = self.SPECIAL_TREATMENT[name](name=name, data=data, description=desc, grid=self.grid, output=self.output, momentumgrid=self.momentumgrid)
             else:
-                o = self.SPECIAL_TREATMENT[name](name=name, data=data, description=desc, grid=self.grid, output=self.output)
+                o = self.SPECIAL_TREATMENT[name](name=name, data=data, description=desc, grid=self.grid, output=self.output, momentumgrid=self.momentumgrid)
```
peterhalldestam, line 115: This assumes that all classes in ``SPECIAL_TREATMENT`` accept the ``momentumgrid`` keyword argument, which is (I believe) why it was added to the ``AvalancheGrowthRate`` and ``LCFSLoss`` constructors in this PR.

If this is the case, then it should also be added in the ``OtherQuantity`` constructor as well. Otherwise, I suspect that there will be a TypeError raised whenever ``f_hot_ripple_pmn`` or ``f_re_ripple_pmn`` is included.

peterhalldestam (changes requested):
There's 2 things that might need to be addressed before a merge. Check out my comments and decide whether the suggested changes are necessary.

py/DREAM/Output/LCFSLoss.py
hoppe93: Good catch! This was a bug since before. Since ``Ire`` is a scalar (in space), giving a radial coordinate makes no sense :P

py/DREAM/Output/OtherQuantities.py
hoppe93: Thanks! You're, absolutely right, I will fix this ;)

peterhalldestam (approved):
Very nice!

Issue #350, opened 2024-10-24
Invalid escape sequences in DREAMIO and DistributionFunction


hoppe93:
In recent versions of Python, the following warnings are reported:
```
DREAM/DREAMIO.py:45: SyntaxWarning: invalid escape sequence '\w'
  m1 = re.search('(\w+://)(.+@)*([\w\-\_\d\.]+)(:[\d]+){0,1}/*(.*)', filename)
DREAM/DREAMIO.py:46: SyntaxWarning: invalid escape sequence '\w'
  m2 = re.search('(.+@)*([\w\-\_\d\.]+):(.*)', filename)
DREAM/Output/DistributionFunction.py:82: SyntaxWarning: invalid escape sequence '\p'
```

PR #351, opened 2024-10-24
Fix issue #350

hoppe93:
Fix warnings reported in newer versions of Python.


peterhalldestam (approved):
Using raw strings is usually considered to be best practice when dealing with regular expressions. Good stuff!

Issue #352, opened 2024-10-24
Error in collisional Redl-Sauter conductivity when Zeff < 1


hoppe93:
Under some circumstances, it seems that Zeff may be evaluated to be < 1 by just a little little bit (probably due to numerical problems), which causes problems for the Redl-Sauter conductivity when collisional effects are included. The problems are due to that ``sqrt(Zeff-1)`` is evaluated explicitly in the code, which will result in a ``NaN`` (or floating-point error in debug mode). 

Since we don't expect ``Zeff < 1``, we should just be able to set ``Zeff = 1`` if it's less than 1 prior to the conductivity computation.

PR #353, opened 2024-10-24
Fix issue #352

hoppe93:
Fix a problem with the collisional Redl-Sauter conductivity when Zeff < 1.


peterhalldestam (approved):
I'm surprised that this can even occur, that ``Zeff`` is less than 1. This same check is already done in ``IonHandler::Rebuild`` where the effective charge number is computed. However, it does not hurt to check this twice, so I will accept this PR.

hoppe93:
Yes, I was also surprised to come across this. It seems to only occur in unconverged iterations (because Zeff >= 1 in the output file), but this was what caused my simulations to crash :P

Issue #345, opened 2024-10-15
Create two separate OtherQuantity objects for radiated power and rate of change in binding energies


peterhalldestam:
In ```Equations/Fluid/RadiatedPowerTerm.cpp``` there is a method that calculates radiation coefficients $L_{\rm line}^{(j)}+L_{\rm free}^{(j)}$ for every ion charge state $j$. This involves line radiation, recombination radiation and bremstrahlung. However, since the change in potential energy due to ionisation and recombination is accounted for in the energy balance equation, the above-mentioned method instead calculates

$$
L_i^{(j)}=L_{\rm line}^{(j)}+L_{\rm free}^{(j)}+\Delta W_i^{(j)}(I_i^{(j)}-R_i^{(j)}).
$$ 

This quantity can be stored as the OtherQuantity object named ```fluid/Tcold_radiation```.

It would be useful to separate these two quantities and, perhaps, let the radiated power remain as simply ```fluid/Tcold_radiation``` and create a new one for the other term, e.g. as ```fluid/Tcold_binding_energy```.

PR #354, opened 2024-10-24
Fix issue #345

peterhalldestam:
Separates the rather misleadingly named OtherQuantity ``Tcold_radiation`` into two terms, namely the radiated power density under the same name ``Tcold_radiation`` and the rate of change in potential energy due to ionisation/recombination under ``Tcold_binding_energy`` (also a power density quantity).


hoppe93 (approved):
Looks great! Well done :)

Issue #167, opened 2022-02-25
Take flux-surface averaged current density as input


hoppe93:
In DREAM, the current density is to be evaluated at the point of minimum B along each flux surface. It is not uncommon in the fusion community to come across current densities which are averaged over flux surfaces. We should therefore implement support in the recently introduced current density-initialization feature of DREAM to give a flux surface-averaged current density, instead of the current density in the point of minimum B along the flux surface.

hoppe93:
One could imagine different averages be given. One possible average, which we should support, is the flux surface average of the toroidal current density, <img src="https://render.githubusercontent.com/render/math?math=\langle\boldsymbol{j}\cdot\nabla\phi\rangle">, which we can express in terms of the parallel current density evolved in DREAM:

<img src="https://render.githubusercontent.com/render/math?math=\left\langle\boldsymbol{j}\cdot\nabla\phi\right\rangle = \frac{j_\parallel}{B}\left[ G\left\langle\frac{1}{R^2}\right\rangle \right]">

Given a flux surface-averaged toroidal current density we could therefore obtain the parallel current density in the point of minimum B by inverting this expression:

<img src="https://render.githubusercontent.com/render/math?math=j_{\parallel,{\rm min}} = \left\langle\boldsymbol{j}\cdot\nabla\phi\right\rangle\frac{B_{\rm min}}{G\left\langle R^{-2} \right\rangle}">

hoppe93:
It turns out that the current density in CORSICA uses a different definition than what was assumed above. The definition used in CORSICA seems to be

$$ j_\phi = \frac{\left\langle\boldsymbol{j}\cdot\boldsymbol{B}\right\rangle}{\left\langle \boldsymbol{B}\cdot\nabla\phi\right\rangle}. $$

This can in turn be rewritten in a form more suitable for DREAM as follows:

$$ \frac{\left\langle\boldsymbol{j}\cdot\boldsymbol{B}\right\rangle}{\left\langle \boldsymbol{B}\cdot\nabla\phi\right\rangle} = \frac{j_{\parallel,\rm min}\left\langle B^2\right\rangle}{B_{\rm in} G\left\langle R^{-2}\right\rangle} $$

which leads to the relation

$$ j_{\parallel,\rm min} = j_\phi\frac{B_{\rm min}G\left\langle R^{-2}\right\rangle}{\left\langle B^2\right\rangle}. $$

PR #355, opened 2024-10-24
Convert between different types of current densities

hoppe93:
This pull request introduces a feature which allows DREAM to automatically convert a given current density profile to the type of parallel current density profile used internally in DREAM. Since there is no general way of handling all possible current density definitions, this pull requests adds support for two types of current densities which have been derived for DREAM. Details can be found in #167.


lvotta98 (approved):
It looks good to me, and the new changes works with my ITER-JOREK benchmark case!

PR #357, opened 2024-10-25
add option 4 to collision handler in DREAM settings

jjwalkowiak:


hoppe93 (approved):
Thanks, this looks good!

PR #358, opened 2024-10-28
Fix issue 275

vaoskar:


hoppe93 (approved):
Great, thanks for fixing this!

PR #364, opened 2024-11-06
Adaptive current gradient-driven transport

hoppe93:
This branch implements a set of new operators for a type of transport which is driven by current density gradients. They are intended as a mitigation for current spikes which form during the CQ of a disruption, in a similar manner to how MHD instabilities are likely driven in experiments to prevent the formation of such spikes.

This branch introduces the following new C++ classes:
- ``RunawayTransportRechesterRosenbluth``: An implementation of the Rechester-Rosenbluth diffusion operator for the ``n_re`` fluid. Now, ``n_re`` transport can be enabled by just specifying $\delta B/B$.
- ``AdaptiveMHDLikeTransportTerm``: An abstract parent class which contains all the routines for determining if the current gradient-driven transport should be activated or not.
- ``AdaptiveHyperresistiveDiffusionTerm``: An implementation of the current gradient-driven transport for the hyperresistive diffusion term.
- ``HeatTransportRRAdaptiveMHDLike``: An implementation of the current gradient-driven transport for the Rechester-Rosenbluth heat diffusion term.
- ``RunawayTransportRRAdaptiveMHDLike``: An implementation of the current gradient-driven transport for the Rechester-Rosenbluth runaway diffusion term.

In addition, the derivation of Rechester-Rosenbluth transport for the runaway density has been documented.

### DIFF
--- a/doc/notes/HyperresistivityFromPerturbation.tex
+++ b/doc/notes/HyperresistivityFromPerturbation.tex
@@ -1,0 +1,92 @@
+\documentclass{notes}
+
+\usepackage[colorlinks=true]{hyperref}
+\usepackage[
+	backend=biber,
+	sorting=none,
+	style=authoryear,
+	citestyle=authoryear
+]{biblatex}
+\addbibresource{ref.bib}
+
+\newcommand{\citep}[1]{(\citename{#1}{author} \citefield{#1}{year})}
+
+\title{Hyperresistivity from magnetic perturbation}
+\author{Mathias Hoppe}
+\date{2024-11-18}
+
+\begin{document}
+	\maketitle
+
+	The thermal quench (TQ) of a disruption is typically associated with a major
+	transport event. In \DREAM, we can apply a number of different types of
+	transport to model such an event. Often we combine heat, ion and runaway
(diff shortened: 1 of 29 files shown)


hoppe93:
We still need to
[ ] Write a common interface for enabling the current gradient-driven transport on ``n_re``, ``psi_p`` and ``T_cold`` in one call.
[ ] Verify that the current implementation of ``AdaptiveMHDLikeTransportTerm`` will work. There is a risk that since we use ``j_tot`` in the current time step to determine whether or not to activate the transport, the enabling of the term might be screwed up (at least with an adaptive time stepper). This can also cause convergence issues potentially. It might be preferable to use the current density in the previous time step for enabling the transport instead.

hoppe93:
- [ ] We also need to ensure that the boundary condition is only set once for transport of ``n_re`` and ``T_cold``.

hoppe93:
- [ ] Add OtherQuantities for the new transport coefficients.

lvotta98 (dismissed):
I leave only one comment, which is more a reminder for the next future.

Everything else looks good in this shape and works properly!

lvotta98 (approved):
Looks good now!

PR #366, opened 2024-11-21
SPI fixes

hoppe93:
A number of fixes to the SPI module, including the Python interface.


py/DREAM/Output/SPIShardPositions.py
```diff
@@ -40,10 +40,15 @@ def arrivalTime(self):
...
 
         # Find the pellet which is travelling the fastest in the x direction
-        imax = np.argmax(np.abs(xp[1,:] - xp[0,:]))
-
-        # Roughly estimate when the fastest pellet reaches the plasma edge
-        it_est = np.argmin(np.abs(RMinusR0[0,-1] - xp[:,imax]))
+        if shard is None:
+            shard = np.argmax(np.abs(xp[1,:] - xp[0,:]))
+            # Roughly estimate when the fastest pellet reaches the plasma edge
+            it_est = np.argmin(np.abs(ROverR0[0,-1] - xp[:,shard]))
```
vaoskar, line 46: Shouldn't it be RMinusR0 instead of ROverR0 here?

py/DREAM/Output/SPIShardPositions.py
```diff
@@ -40,10 +40,15 @@ def arrivalTime(self):
...
+        else:
+            xp = xp[:,shard].reshape((xp.shape[0], 1))
+            yp = yp[:,shard].reshape((yp.shape[0], 1))
+            # Roughly estimate when the fastest pellet reaches the plasma edge
+            it_est = np.argmin(np.abs(ROverR0[0,-1] - xp[:,0]))
```
vaoskar, line 51: Same as above

vaoskar (changes requested):
Looks good! There were just a few minor things that I was wondering about that may be good to check before merging

vaoskar (approved):

Issue #367, opened 2024-11-21
Inconsistent initialization of electric field-dependent quantities when E_field is initialized from j_tot


hoppe93:
It turns out that when ``E_field`` is initialized from ``j_tot`` (i.e. a current density profile is prescribed), and ``f_re`` is initialized according to an avalanche distribution, the distribution function will end up initialized with ``E_field=0``. The cause for this is the fact that to avoid a circular dependency, ``E_field`` must be initialized to 0 at the start of the initialization cycle, after which ``RunawayFluid`` can be initialized, followed again by an initialization of ``E_field`` from ``j_tot`` and the conductivity (contained in ``RunawayFluid``).

Since ``E_field`` will be initialized at the start of the initialization, and since ``j_tot`` depends on ``j_re`` (and thereby ``f_re``), ``f_re`` will be initialized with ``E_field=0`` and therefore become zero.

A better approach to initializing ``f_re`` would be to allow the user to provide the ``E_field`` profile to use. Yet another alternative could be to separate the conductivity from ``RunawayFluid``, as suggested in #84.

PR #368, opened 2024-11-21
Add parameter 'E_init' for f_re initialization

hoppe93:
Resolves issue #367 by adding a new input parameter ``E_init`` for initializing the runaway distribution function.

### DIFF
--- a/src/Settings/Equations/f_re.cpp
+++ b/src/Settings/Equations/f_re.cpp
@@ -40,6 +40,8 @@ void SimulationGenerator::DefineOptions_f_re(Settings *s) {
 
     s->DefineSetting(MODULENAME "/inittype", "Specifies how to initialize f_re from n_re.", (int_t)OptionConstants::UQTY_F_RE_INIT_FORWARD);
 	//s->DefineSetting(MODULENAME "/initavag0", "Gamma0 parameter in analytical avalanche distribution when initializing using this distribution.", (real_t)20.0);
+	
+	DefineDataR(MODULENAME, s, "/E_init");
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/Settings/Equations/f_re.cpp
```diff
@@ -188,7 +188,7 @@ void SimulationGenerator::ConstructEquation_f_re_kineq(
                             // will in general be different from the quantity n_re
                             for (len_t j = 0; j < np2; j++)
                                 for (len_t i = 0; i < np1; i++)
-                                    finit[offset + j*np1 + i] = fRE->evaluateFullDistribution(ir, xi[j], p[i]);
+                                    finit[offset + j*np1 + i] = fRE->evaluateFullDistributionWithE(ir, xi[j], p[i], E_init[ir]);
```
IdaEkmark, line 191: Now the function "evaluateFullDistribution" is never used anywhere, if I'm not mistaken, but that is probably fine

examples/init_f_re/generate.py
IdaEkmark, line 1: To make this script a bit more pedagogical, could we not use $E\neq1$ (e.g. $E=1.5$), and give that $E$ as an argument to ```setInitialAvalancheDistribution```? Right now it feels very convinient that $E = 1$ and the default value of $E_{\rm init}$ is 1

IdaEkmark (commented):
One small suggestion, otherwise, I'm happy to approve

src/Settings/Equations/f_re.cpp
hoppe93: Yeah, I think it's still a relevant feature to have. Especially since the original idea with the ``AnalyticDistribution`` classes was to enable them to substitute numerical distribution function. And there are occasionally still discussions about completing this work...

examples/init_f_re/generate.py
hoppe93: Great idea! I've implemented this now :)

IdaEkmark (approved):
Now I'm happy!

PR #370, opened 2024-11-25
Small change in string interpreting for FluidQuantity plotting

fusionby2030:
A syntax warning was being thrown with invalid escape sequence `\l`, thererefore cast the string as a Raw string. Typically done for latex in matplotlib (e.g., `'\alpha'` requires `r'\alpha'`). 

The warning was thrown everytime `DREAMpy` was imported. I disabled it locally, but thought to fix it for other new users.


hoppe93 (approved):
Many thanks for this, Adam! It looks good, and I will merge it as soon as the tests are finished :)

PR #371, opened 2024-11-26
Small update and bug fix

IdaEkmark:
The Moment-function of the KineticQuantity-class always used Vprime/VpVol w.r.t. r=0, instead of r=iR, which we have fixed. We also added toe functionality to plot negative quantities with logarithmic scales.


hoppe93 (changes requested):
Nice! I just have two thoughts which we can either fix or just ignore and merge.

hoppe93 (approved):
Great!

PR #378, opened 2024-11-28
Add ionization time as OtherQuantity

hoppe93:
Add the ionization time scale (as calculated in the adaptive ionization time stepper) as an other quantity.

### DIFF
--- a/src/OtherQuantityHandler.cpp
+++ b/src/OtherQuantityHandler.cpp
@@ -462,6 +462,14 @@ void OtherQuantityHandler::DefineQuantities() {
                 vec[ir] = 0;
             this->tracked_terms->T_cold_ohmic->SetVectorElements(vec, Eterm);
         );
+	if (tracked_terms->T_cold_halo != nullptr)
+        DEF_FL("fluid/Tcold_halo", "Power density escaping through the halo region [J s^-1 m^-3]",
+            real_t *Wcold = this->unknowns->GetUnknownData(this->id_Wcold);
+            real_t *vec = qd->StoreEmpty();
+            for(len_t ir=0; ir<this->fluidGrid->GetNr(); ir++)
+                vec[ir] = 0;
+            this->tracked_terms->T_cold_halo->SetVectorElements(vec, Wcold);
+		);
     if (tracked_terms->T_cold_fhot_coll != nullptr)
         DEF_FL("fluid/Tcold_fhot_coll", "Collisional heating power density by f_hot [J s^-1 m^-3]",
             real_t *fhot = this->unknowns->GetUnknownData(id_f_hot);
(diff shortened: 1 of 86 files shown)


src/OtherQuantityHandler.cpp
```diff
@@ -430,6 +432,22 @@ void OtherQuantityHandler::DefineQuantities() {
     DEF_FL("fluid/tauEERel", "Relativistic electron collision time (4*pi*lnL*n_cold*r^2*c)^-1 [s]", qd->Store(this->REFluid->GetElectronCollisionTimeRelativistic()););
     DEF_FL("fluid/tauEETh", "Thermal electron collision time (tauEERel * [2T/mc^2]^1.5) [s]", qd->Store(this->REFluid->GetElectronCollisionTimeThermal()););
 
+	DEF_FL_T("fluid/tIoniz", "Estimate of the ionization time scale (as used by the adaptive ionization time stepper) [s]",
+		real_t *n = this->unknowns->GetUnknownData(this->id_ncold);
```
peterhalldestam, line 436: When trying to run with this new oqty the example script ``examples/ts_ioniz/generate.py``, I can indeed find ``tIoniz``, but all values are set to ``inf`` while ``n_cold`` is changing in time.

For some reason, the two methods ``GetUnknownData`` and ``GetUnknownDataPrevious`` are yielding the same values on line ``src/OtherQuantityHandler.cpp:436`` and onward. Also the difference in time is zero...

peterhalldestam (changes requested):
See whether you can reproduce the issue I had when running the example script ``examples/ts_ioniz/generate.py``.

src/OtherQuantityHandler.cpp
```diff
@@ -430,6 +432,22 @@ void OtherQuantityHandler::DefineQuantities() {
...
+		real_t *np = this->unknowns->GetUnknownDataPrevious(this->id_ncold, 1);
+		real_t tp = this->unknowns->GetUnknownDataPreviousTime(this->id_ncold, 1);
```
peterhalldestam, line 438: When printing ``tp`` and all elements of ``np`` every timestep, it seems like they are always equal to 0, for some reason. Since ``np`` is 0, the ionisation time is therefore 0 too

peterhalldestam (changes requested):
Still not working as intended. Instead of the ionisation time is set to infinity, now it is 0 for some reason. There's still problem with retrieving data and time from the previous step

src/OtherQuantityHandler.cpp
peterhalldestam: Also, please add it to the list of other quantities in ``py/DREAM/Settings/OtherQuantities.py``

peterhalldestam: I realise now that I get this issue when running ``examples/pyface-terminate/generate.py``, which uses dreampyface to run the simulation. This is however not an issue when running "examples/ts_ioniz/generate.py", which uses the good ol' runiface.

peterhalldestam (approved):

PR #389, opened 2024-12-12
Changed pprime from interpolation to actual data

IdaEkmark:
In EQDSK tool, changed $p'$ so that it is given by the data rather than interpolated from the data for the pressure.


hoppe93 (approved):
Excellent!

PR #391, opened 2024-12-17
Code information in output

hoppe93:
This pull request adds a number of variables to the output file which can help when reproducing DREAM output.


lvotta98 (approved):
Looks good!

Issue #392, opened 2024-12-19
Custom radial grid is broken


hoppe93:
A recent commit broke the function of the custom radial grid, since the C++ code checks for ``nr=0`` to determine whether to use the custom radial grid or not.

PR #393, opened 2024-12-19
Fix issue #392

hoppe93:
Fix the issue reported in #392, which reported on the custom radial grid no longer working.


lvotta98 (approved):
Now it works well!

Issue #394, opened 2025-01-11
Bug when reading NumPy arrays in C++


hoppe93:
A bug has been identified in ``dreampy_convert()`` of ``dreampyface/cxx/SFile_Python.cpp``. This function currently assumes that input arrays are contiguous, which is not always the case. For some reason, in one of @peterhalldestam 's simulations, the ``Y_p`` vector is a non-contiguous array when passed to the C++ code, with a stride of 32. Normal arrays are contiguous with a stride of 8. Hence, when read in the C++ code (which does not account for the stride), garbage data appears.

So, we need to account for the stride parameter of the NumPy array (see https://numpy.org/doc/stable/reference/c-api/array.html) in the C++ code.

PR #395, opened 2025-01-13
Fix issue #394

hoppe93:
This pull request resolves a bug in dreampyface related to NumPy arrays with unusual strides.


peterhalldestam (approved):
Works like a charm!

PR #396, opened 2025-01-13
Fix tritium RE bug

lvotta98:
This pull request fix a bug in the python interface which always disabled RE generation from Tritium.


hoppe93 (approved):
This looks good to both me and Ida!

Issue #401, opened 2025-01-31
Wrong initialization of avalanche distribution


hoppe93:
The initial avalanche distribution for ``f_re`` should not quite be initialized as now. Instead of evaluating both the energy and pitch distributions with a provided ``E_init`` value, only the pitch distribution should be initialized as such. The electric field is supposed to cancel the same factor in the avalanche growth, which appears in the energy distribution, and so the same electric field has to be used both as input to the ``AnalyticDistributionRE`` and the avalanche growth rate (which is evaluted in ``RunawayFluid`` and therefore always uses the current value for ``E_field``).

PR #402, opened 2025-01-31
Fix issue #401

hoppe93:
This pull request implements a fix for #401. Since ``E-Eceff`` in the exponent of the energy distribution is supposed to cancel the same factor in the avalanche growth rate (which is evaluated in ``RunawayFluid``, and therefore always uses the latest value for ``E_field``), the energy distribution should always be evaluated with the latest value for ``E_field``, regardless of what value is used for the pitch distribution.


IdaEkmark (approved):
Looks good!

PR #405, opened 2025-02-03
Add kinetic ionization rate as other quantity

hoppe93:
Add the runaway ionization rate (density per unit time) as other quantities.


peterhalldestam (approved):
Nice!

PR #408, opened 2025-03-17
Use expint from GSL

BjornZaar:
DREAM failed to build with Clang, as "expint" in SPIHandler.cpp was not recognized.

Changed "expint" to "gsl_sf_expint_Ei" instead .


hoppe93:
I re-ran all our ITER simulations with this fix and confirmed that it yields identical results to before this fix.

Results: http://130.237.45.201/run/14

hoppe93 (approved):
Looks good and yields identical results to before the fix.

Issue #411, opened 2025-03-31
Memory segmentation fault with ds.eqsys.n_i.addIonSource()


ericend:
Memory segmentation fault while running simulation of thermal quench. Tries to delete the source strength term dNdt twice.

hoppe93:
Somewhat unrelated to this, but should also be fixed in the same pull request: information about the ion source is not loaded in the ``fromdict()`` method of ``Ions.py``, which means that settings are not retained when copying/loading ``DREAMSettings`` objects.

PR #412, opened 2025-03-31
Fix bug in IonSourceTerm

hoppe93:
As reported in #411, the ion source term halts with a segmentation fault. The issue was due to that for each species, one ``IonBoundaryCondition`` is created and given a reference to a single ``MultiInterpolator1D`` object representing the ion source term prescription. When the code finishes, it deletes each of the ``IonBoundaryCondition``s, which in turn calls the destructor for ``MultiInterpolator1D``. When this is done for the second time, the ``MultiInterpolator1D`` was already deleted when the destructor is called, which results in a segmentation fault.

The solution to this problem was to load one ``MultiInterpolator1D`` for each ``IonBoundaryCondition``. This may consume unnecessary memory, but simplifies the code (and does not consume excessive amounts of memory since the ``MultiInterpolator1D`` occupies memory of order O(nr*nIons)).

### DIFF
--- a/py/DREAM/Settings/Equations/Ions.py
+++ b/py/DREAM/Settings/Equations/Ions.py
@@ -4,7 +4,7 @@ import numpy as np
 import scipy.interpolate
 from numpy.matlib import repmat
 from DREAM.Settings.Equations.EquationException import EquationException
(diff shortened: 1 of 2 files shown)


ericend:
Recompiled on branch iss411 and the problem seems to be solved!

lvotta98 (approved):
This looks good to me!

PR #416, opened 2025-04-03
Removed the 6.29/2 correction constant in dBB calculation

### LINKED ISSUE #413: Correction constant 6.29/2 in TransportCoefficientReader.py must be removed
There is a factor 6.29/2 that was added to line 121 in `TransportCoefficientReader.py` to compensate for a previous error in the provided value for the plasma major radius in ITER disruption simulations. Adding this factor does gives the same `dB/B` for the heat transport as in previous simulations, which is what we want in order to get the desired temperature drop with transport as the only loss mechanism. However, the diffusion coefficient for the REs will not be consistent with the one for the heat. Instead, a better strategy is to remove this correction factor from `TransportCoefficientReader.py`  and re-scale the  `t_duration_over_t_diffusion` with a factor 6.29/2 which is specific for ITER simulations, getting both the correct `dB/B` for the heat transport and a consistent RE diffusion coefficient.

lvotta98:


hoppe93 (changes requested):
I think the comment just above this line (and at the end of the line) should be removed as well. As it is now it is quite confusing to keep them.

hoppe93 (dismissed):

hoppe93 (approved):
Looks good!

PR #417, opened 2025-04-07
Runaway ionization scale factor

hoppe93:
This pull request adds a scaling factor to the runaway ionization source term.


hoppe93:
Something should also be added to the documentation about this factor.

peterhalldestam (approved):

PR #420, opened 2025-04-14
NBIHeatTerm

ekmanlinn:
# Implementation of Neutral Beam Injection (NBI) Heating Term

## Description
Adds a new equation term in for the cold electron temperature calculation, handling electron heating from neutral beam injection. The implementation includes a new class `NBIElectronHeatTerm` for calculating the NBI heating contribution.

## Also includes
- Support for arbitrary beam injection geometry
- Visualization tools for beam-plasma geometry

## Implementation Details
- Implements beam-aligned coordinate system (s, r, θ) for efficient beam tracing
- Uses precomputed orthonormal basis vectors (n_norm, e1, e2) for coordinate transformations
- Caches flux surface geometry for performance
- Handles arbitrary beam injection angles and entry points

- Ionization processes through ADAS rate coefficients
  - Uses SCD coefficients for electron impact ionization
- Charge exchange reactions
  - Implements CCD coefficients for charge exchange with plasma ions
- Beam attenuation
  - Calculates survival probability along beam paths
  - Accounts for both ionization and charge exchange losses
  - Uses adaptive step size for integration

- Volume integration over flux surfaces
  - Proper Jacobian for toroidal geometry
  - Handles flux surface geometry through cached interpolation

- Beam current profile
  - Supports arbitrary j_B(r) profiles through interpolation
  - Normalizes to total beam current via numerical integration


hoppe93 (changes requested):
Great work! Here's a first batch of comments for you to have a look at!

examples/nbi_test/generate.py
```diff
@@ -0,0 +1,235 @@
...
+
+# Initial current / E-field setup
+ds.eqsys.j_ohm.setInitialProfile(1, Ip0=Ip_initial)
+ds.eqsys.E_field.setType(EField.TYPE_SELFCONSISTENT)
+ds.eqsys.E_field.setBoundaryCondition(EField.BC_TYPE_PRESCRIBED, V_loop_wall_R0=0, R0=R0)
+
+# Set up electron temperature and NBI
+ds.eqsys.T_cold.setType(T_cold.TYPE_SELFCONSISTENT)
+ds.eqsys.T_cold.setInitialProfile(T_initial)
+ds.eqsys.T_cold.include_NBI = True
```
hoppe93, line 76: It would be nicer if this is set automatically when ``setNBI()`` is called.

hoppe93 (changes requested):
Here come a few more comments, this time on the ``NBIElectronHeatTerm`` cpp files.

cmake/FindPETSc.cmake
```diff
@@ -27,6 +27,10 @@ cmake_policy(VERSION 3.12)
 set(PETSC_VALID_COMPONENTS
   C
   CXX)
+if(PETSC_INCLUDES AND PETSC_LIBRARIES)
+    set(PETSC_FOUND TRUE)
+    return()
+endif()
```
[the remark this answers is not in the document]
ekmanlinn: I have to keep this, otherwise doesnt work on my computer. As you said we can figure out why another time

src/OtherQuantityHandler.cpp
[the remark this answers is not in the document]
ekmanlinn: Is it really neccesary since its just a linear scalar to the NBIElectronHeatTerm?

examples/nbi_test/generate.py
ekmanlinn: Does that work? From my understanding the NBI is only enabled for TYPE_SELFCONSISTENT so doesnt that have to me "turned on" first.

src/Equations/Fluid/NBIElectronHeatTerm.cpp
```diff
@@ -0,0 +1,619 @@
...
+        H_r_dni[ir] = 0.0;
+        H_r_dTi[ir] = 0.0;
+        H_r_dne[ir] = 0.0;
+    }
+    std::vector<real_t> lambda_cache(nr, -1.0); // -1 = not yet computed
+    std::vector<real_t> dlambda_dI_cache(nr, -1.0);
+    std::vector<real_t> dlambda_dne_cache(nr, -1.0);
+
+    for (real_t i_beam_theta = 0; i_beam_theta < n_beam_theta; i_beam_theta++)
+    {
```
hoppe93, line 498: This curly brace should be at the end of the line above.

hoppe93 (changes requested):
Great work! I have some comments, although mainly regarding the code formatting.

PR #424, opened 2025-04-22
Fixes typo in docs for tau_T

BjornZaar:
There was a typo in docs/frontend/api/Settings/eqsys/n_re where the tritium half-life was given as 4800 d instead of 4500 d.


hoppe93 (approved):
Looks good! Thanks for fixing!

Issue #425, opened 2025-04-27
Bug in transport operator verification


hoppe93:
It appears that line 667 in ``TransportSettings.py`` is not working as intended, as the if-statement is supposed to guard against a ``KeyError`` when calling ``g()``.

PR #426, opened 2025-04-27
Fix issue #425

hoppe93:
Fix typo on line 667 in ``TransportSettings.py``.


IdaEkmark (changes requested):
I might be missing something, but I think the fix is not doing what it intends to do

IdaEkmark (approved):
Looks great!

PR #427, opened 2025-05-27
Re-implemented the Kiramov BC

IdaEkmark:
Re-implemented the Kiramov BC, but adapted it to be consistent with the HaloCurrentRegion-class


src/Settings/Transport.cpp
```diff
@@ -599,6 +600,15 @@ bool SimulationGenerator::ConstructTransportTerm(
...
 		}
 	}
+
+    if (heat && bc==OptionConstants::EQTERM_TRANSPORT_BC_KIRAMOV){
+
+        TransportAdvectiveBC * t = new KiramovBoundaryHeatTransportBC(grid, eqsys->GetUnknownHandler(), eqsys->GetIonHandler());
+        oprtr->AddBoundaryCondition(t);  
+
+        if (advective_bc != nullptr)
+            *advective_bc = t;
```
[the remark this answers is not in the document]
IdaEkmark: Since we only need the BC if we have (non-trivial) transport, it should be fine to just add `hasNonTrivialTransport` to the if-statement on line 604, right?

src/Equations/Fluid/KiramovBoundaryHeatTransport.cpp
```diff
@@ -0,0 +1,148 @@
...
+	// case as we should only have 1 on BC in the radial direction)
+    real_t q_par = 2 * kappa * n_e * c_s * Constants::ec;
+
+    Fr(nr,0,0) = 2. / 3. * q_par / L_par * 1. / S_wo_coeff; 
+}
+
+// Set derivatives of the advection term with respect to different unknowns quantities 
+void KiramovBoundaryHeatTransport::SetPartialAdvectionTerm(len_t derivId, len_t /*nMultiples*/){ 
+    if(GetAdvectionDiffCoeffR() == nullptr)
+        AllocateDifferentiationCoefficients();
```
IdaEkmark, line 108: The differentiation coefficients are never allocated, therefore I put it here. For a non-BC advection term, the coefficients would be located in `AdvectionTerm::SetJacobianBlock`. For an advective BC, `DREAM::TransportBC<T>::AddToJacobianBlock` is used instead, which follows the same structure as `AdvectionTerm::SetJacobianBlock`, except for not having anything corresponding to  `AllocateDifferentiationCoefficients()`. 

Is this the best location for `AllocateDifferentiationCoefficients()`? Is it better if `DREAM::TransportBC<T>::AddToJacobianBlock` is updated to allocate instead?

hoppe93: I think ``SetPartialAdvectionTerm()`` sounds like a reasonable place, since it should be the first place where those coefficients are used within ``KiramovBoundaryHeatTransport``. I think allocating the data in ``AddToJacobianBlock()`` might be too late (unless you also move the rebuild logic there, which I think goes against the purpose of that method).

hoppe93: Are these coefficients deallocated somewhere as well?

IdaEkmark: `AddToJacobianBlock` is higher up in the chain of function calls, i.e. `SetPartialAdvectionTerm` is called inside of `AddToJacobianBlock` (`DREAM::TransportBC<T>::AddToJacobianBlock` calls `DREAM::TransportBC<DREAM::FVM::AdvectionTerm>::SetPartialTerm` calls `KiramovBoundaryHeatTransport::SetPartialAdvectionTerm`), so I don't think adding it in `AddToJacobianBlock` would be too late

IdaEkmark: > Are these coefficients deallocated somewhere as well?

`DeallocateDifferentiationCoefficients` is called in the destructor of `AdvectionTerm`, and since `KiramovBoundaryHeatTransport` inherits from `AdvectionTerm`, deallocation is handled correctly as long as `KiramovBoundaryHeatTransport` is deleted, right? In that case, I think I have fixed it now.

hoppe93 (approved):
Thanks for implementing this! Looks good to me now! :)

PR #431, opened 2025-06-09
Fixed bug

IdaEkmark:
I fixed issue 430, but I don't know if it's the best implementation, and maybe we want to add something more.


hoppe93 (approved):
This looks good now! :)

PR #433, opened 2025-06-13
Initial wall current

hoppe93:
This branch implements a setting for the initial wall current. So far, the initial wall current has always been set to zero, but there may situations in which one would like a non-zero initial wall current.


lvotta98 (approved):
It looks good to me!

PR #436, opened 2025-07-02
updated setup files

peterhalldestam:
Just changed the names of the module that are imported for the updated TOK cluster. Still there's some issue with running libdreampy...

### DIFF
--- a/setup/build.tok.sh
+++ b/setup/build.tok.sh
@@ -21,7 +21,7 @@
 #
 # 2024-03-25
 # - Minor fix, it can now be installed on TOK machines. Does not seem to work with MKL, however... /Peter
-
+#
 source environment.tok.sh
 
 function install_petsc {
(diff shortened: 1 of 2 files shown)


setup/build.tok.sh
```diff
@@ -41,9 +41,11 @@ function install_petsc {
 }
 function install_dream {
 	cd "$DREAMPATH" && rm -rf build && mkdir build && cd build &&
-	cmake .. -DPETSC_EXECUTABLE_RUNS=YES -DPETSC_DIR=$PETSC_DIR -DPETSC_ARCH=$PETSC_ARCH &&
+	cmake .. -DDREAM_BUILD_PYFACE=OFF -DPETSC_EXECUTABLE_RUNS=YES -DPETSC_DIR=$PETSC_DIR -DPETSC_ARCH=$PETSC_ARCH
```
hoppe93, line 44: Is the ``DREAM_BUILD_PYFACE=OFF`` an active choice, or is it required to avoid other issues?

hoppe93 (approved):
This looks good to me! :)

setup/build.tok.sh
peterhalldestam: I got an error because the Python environment (python-waterboa) is loading an older version of libstdc++.so.6, but one of the libraries (likely libhdf5_cpp.so.310, built with GCC 14) needs a newer C++ ABI (CXXABI_1.3.15). This mismatch seems to cause a runtime failure when loading shared libraries. 

There should be a simple workaround to this, I'm sure. But for now, I'll leave it like this...

Issue #438, opened 2025-07-09
NumPy array tostring() removed


hoppe93:
The method ``numpy.ndarray.tostring()`` has been removed in recent versions of NumPy, after being deprecated since version 1.19.0. Instead of this method, ``numpy.ndarray.tobytes()`` should be used.

PR #439, opened 2025-07-09
Fix issue #438

hoppe93:
Replaces calls to the now removed ``numpy.ndarray.tostring()`` method with calls to ``numpy.ndarray.tobytes()``.


IdaEkmark (approved):
Looks good!

PR #442, opened 2025-07-16
Iss440

IdaEkmark:


src/Equations/Fluid/LCFSLossRateTerm.cpp
```diff
@@ -103,6 +103,12 @@ void LCFSLossRateTerm::FindRadiusOfLCFS(){
     if ( exists == false ) { this->ir_LCFS = -1; } // No flux surface is closed
 }
 
+real_t LCFSLossRateTerm::GetRadiusOfLCFS() {
+    FindRadiusOfLCFS();
```
IdaEkmark, line 107: Not sure if this is needed

hoppe93: It should not be needed, as it seems `FindRadiusOfLCFS()` is called on every rebuild (and the other quantity is always evaluated after this).

hoppe93 (approved):
The implementation looks OK to me. I added some comments which you might want to have a look at before merging (especially the softlib comment; it's fine to upgrade the version, but please don't downgrade).

hoppe93 (approved):
Looks good!

PR #445, opened 2025-08-04
Tool for saving volume integrated distribution function in file

IdaEkmark:
Will be relevant for using DREAM-output data for wall load calculations, e.g. using HEAT


tools/saveDistributionFunction.py
```diff
@@ -0,0 +1,81 @@
...
+    parser = argparse.ArgumentParser()
+
+    # Arguments
+    parser.add_argument('output', help="Name of DREAM output file (excluding .h5)", action='store', default='', type=str)
+    parser.add_argument('filename', help="Name of file to save data in (excluding file extension)", action='store', default='', type=str)
+
+    # Options
+    parser.add_argument('-E', '--energy', help="Use energy pitch angle coordinates", action='store_const', const="E theta", dest="coordinates", default="p xi")
+    parser.add_argument('-C', '--cylindrical', help="Use cylindrical coordinates", action='store_const',const="p_par p_perp", dest="coordinates", default="p xi")
+    parser.add_argument('-M', '--matfile', help="Use mat file format to save data", action='store_true', dest="mat", default=False)
```
hoppe93, line 19: The help text could be more specific. The format is "Matlab version 5".

hoppe93 (changes requested):
Looks good! I just have a few small comments.

tools/saveDistributionFunction.py
```diff
@@ -0,0 +1,82 @@
+import numpy as np
+import scipy.constants as const
+from scipy.io import savemat
+import sys, argparse, h5py
+sys.path.append('../py')
```
hoppe93, line 5: I think we should change this to
```python
path = str((pathlib.Path(__file__).parent / '../py').resolve().absolute())
sys.path.append(path)
```
This will also require ``pathlib``:
```python
import pathlib
```

tools/saveDistributionFunction.py
```diff
@@ -0,0 +1,82 @@
...
+
+    if settings.mat:
+        savemat(f"{settings.filename}.mat", dic)
+    else:
+        with h5py.File(f"{settings.filename}.h5", 'w') as hf:
+            for key, value in dic.items():
+                if isinstance(value, (np.ndarray, list, tuple)):
+                    hf.create_dataset("/" + key, data=np.array(value))
+                else:
+                    hf.create_dataset("/" + key, data=value)
```
hoppe93, line 77: These can be written as
```python
hf[key] = np.array(value)
```
and
```python
hf[key] = value
```
respectively, instead.

hoppe93 (changes requested):
Just a few more fixes!

PR #456, opened 2025-11-07
NBI Handler

ekmanlinn:
# NBI Implementation 

This pull request implements neutral beam injection (NBI) heating in DREAM, including both electron and ion heating.

## NBI Heating Implementation (`NBIHandler`)

1. **Radial Power Deposition Calculation**
   - Calculates NBI power deposition `H(r)` as a function of radius
   - Supports multi-energy beam components (full, half, third energy)
   - Returns radially-resolved power density `[W/m³]` at each flux surface

2. **Electron Heating (`Tcold_NBI` term)**
   - A fraction `f_e(r)` of the deposited power heats electrons
   - Creates `NBIElectronHeatingTerm` which inherits from `EquationTerm`
   - Added to the `T_cold` (cold electron temperature) equation
   
3. **Ion Heating (`Wi_NBI` term)**
   - Remaining fraction `f_i(r) = 1 - f_e(r)` heats ions
   - Creates `NBIIonHeatingTerm` which inherits from `IonEquationTerm`
   - Added to the `W_i` (ion energy) equation for each ion species

#### Spatially Uniform Ion Source (`ION_SOURCE_PRESCRIBED_VOLUMETRIC`)

Implemented a new ion source type that allows  injection rate over the radial grid.

**Note:** The `nbi_test/` directory contains testing/validation scripts that were are used during testing but are not critical for code review.

**Note:** The file NBIHandler.cpp is very similar to the old NBIElectronHeat.cpp that was mearched before.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,5 @@ petsc_matrix
 petsc_matrix.info
 src/Atomics/adasdata.cpp
 tools/cache
+*.CDF
+*.cdf
--- a/fvm/Grid/AnalyticBRadialGridGenerator.cpp
+++ b/fvm/Grid/AnalyticBRadialGridGenerator.cpp
@@ -751,8 +751,12 @@ const real_t *AnalyticBRadialGridGenerator::GetFluxSurfaceRMinusR0() {
 	for (len_t j = 0, i = 0; j < ntheta; j++) {
 		real_t theta = 2*M_PI*j / ntheta;
 
-		for (len_t ir = 0; ir < nr; ir++, i++)
-			R[i] = ROverR0AtTheta(ir, theta) * this->R0 - this->R0;
+		for (len_t ir = 0; ir < nr; ir++, i++) {
+			if (std::isinf(this->R0))
+				R[i] = r[i] * cos(theta);
+			else
+				R[i] = ROverR0AtTheta(ir, theta) * this->R0 - this->R0;
+		}
 	}
 
 	return R;
@@ -772,7 +776,10 @@ const real_t *AnalyticBRadialGridGenerator::GetFluxSurfaceRMinusR0_f() {
 		real_t theta = 2*M_PI*j / ntheta;
 
 		for (len_t ir = 0; ir < nr+1; ir++, i++)
-			R[i] = ROverR0AtTheta_f(ir, theta) * this->R0 - this->R0;
+			if (std::isinf(this->R0))
+				R[i] = r_f[i] * cos(theta);
+			else
+				R[i] = ROverR0AtTheta_f(ir, theta) * this->R0 - this->R0;
 	}
 
 	return R;
@@ -782,7 +789,10 @@ const real_t *AnalyticBRadialGridGenerator::GetFluxSurfaceRMinusR0_f() {
  * not shifted by R0.
  */
 real_t AnalyticBRadialGridGenerator::GetFluxSurfaceRMinusR0_theta(len_t ir, real_t theta){
-    return ROverR0AtTheta_f(ir, theta) * this->R0 - this->R0;
+	if (std::isinf(this->R0))
+		return r_f[ir] * cos(theta);
+	else
+		return ROverR0AtTheta_f(ir, theta) * this->R0 - this->R0;
 }
  
 
@@ -790,7 +800,6 @@ real_t AnalyticBRadialGridGenerator::GetFluxSurfaceRMinusR0_theta(len_t ir, real
  * Returns the flux surface Z coordinates on the simulation grid edges.
  */
 real_t AnalyticBRadialGridGenerator::GetFluxSurfaceZMinusZ0_theta(len_t ir, real_t theta){
-
     return this->r_f[ir] * this->kappa_f[ir] * sin(theta);
 }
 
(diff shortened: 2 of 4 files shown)


include/DREAM/Settings/OptionConstants.enum.hpp
```diff
@@ -63,7 +63,8 @@ enum ion_neutral_advection_mode {
 
 enum ion_source_type {
 	ION_SOURCE_NONE=1,
-	ION_SOURCE_PRESCRIBED=2
+	ION_SOURCE_PRESCRIBED=2,
+	ION_SOURCE_PRESCRIBED_CONSTANT=3
```
hoppe93: What does the ``CONSTANT`` refer to here? I think it would be better to call it ``ION_SOURCE_PRESCRIBED_VOLUMETRIC`` or so, since it shouldn't be constant in time nor space.

src/Equations/Fluid/NBIElectronTerm.cpp
```diff
@@ -0,0 +1,183 @@
...
+ * Rebuild the term (called once per time step). Call of the build function in NBIHandler
+ */
+void NBIElectronTerm::Rebuild(const real_t t, const real_t dt, FVM::UnknownQuantityHandler* unknowns) {
+    real_t beam_energy = handler->GetBeamEnergy();
+
+    // Helper to copy data
+    auto copyData = [&](real_t energy, real_t* Qe_target, real_t* d_Te_target,
+                        real_t* d_ne_target, real_t* d_n_ij_target,
+                        real_t* d_T_ij_target) {
+        handler->Build(t, dt, unknowns, energy);
```
hoppe93: Should this method really be called here? This way, it will be called three times within ``NBIElectronTerm::Rebuild()``, seemingly with identical input parameters each time. For example the ``RunawayFluid`` and ``SPIHandler`` classes are built in ``src/Solver/Solver.cpp``, in the ``RebuildTerms()`` method, which is called exactly once per Newton iteration. There, you can also find infrastructure for measuring the time spent re-building the module.

src/Equations/Fluid/NBIIonTerm.cpp
```diff
@@ -0,0 +1,274 @@
...
+    const real_t* energy_fractions = handler->GetEnergyFractions();
+    
+    len_t NZ = ions->GetNZ(); 
+    if (NZ == 0) {
+        throw DREAMException("No ion species found in NBIIonTerm");
+    }
+    std::vector<real_t> ni_species(NZ, 0.0);   
+    std::vector<real_t> Zi(NZ, 0.0);           
+    std::vector<real_t> Mi(NZ, 0.0);  
+    if (Z0 != 0) return;
```
hoppe93: Why not return _before_ anything else in this method?

src/Equations/Fluid/NBIIonTerm.cpp
```diff
@@ -0,0 +1,274 @@
...
+    
+
+/**
+ * Set the matrix elements corresponding to this term.
+ */
+void NBIIonTerm::SetCSMatrixElements(
+            FVM::Matrix *mat, real_t *rhs,
+        const len_t iIon, const len_t Z0, const len_t /*rOffset*/
+){
+    (void)mat;
```
hoppe93: Why?

src/Equations/Fluid/NBIIonTerm.cpp
```diff
@@ -0,0 +1,274 @@
...
+            len_t col = ir;
+            real_t deriv = energy_fractions[0] * d_Qe1_d_ne[ir]
+                         + energy_fractions[1] * d_Qe2_d_ne[ir]
+                         + energy_fractions[2] * d_Qe3_d_ne[ir];
+            real_t w= ComputeWeightFactor(ir, iIon);
+           jac->SetElement(row, col, w*deriv); 
+        }
+    }
+    
+    if (derivId == id_Tcold) {
```
hoppe93: ``else if``?

src/NBIHandler.cpp
```diff
@@ -0,0 +1,757 @@
...
+    d_NBIHeatTerm_e_d_n_ij = new real_t[maxidx]();
+    H_r_dn_ij = new real_t[maxidx]();
+    H_r_dT_ij = new real_t[maxidx]();
+}
+
+/**
+ * Configure the NBIHandler from NBI settings. Called at the start of the simulation.
+ */
+void NBIHandler::ConfigureFromSettings(
+    Settings* /*s*/, FVM::UnknownQuantityHandler* unknowns, real_t s_max, real_t r_beam,
```
hoppe93: If you are not using a parameter in the argument list you can just leave it unnamed. Sometimes we put the name in a comment when the name can help in identifying what the meaning of the parameter is. In this case, the type of ``Settings *s`` is sufficiently clear to not require the variable name to be kept.

src/NBIHandler.cpp
```diff
@@ -0,0 +1,757 @@
...
+                d_NBIHeatTerm_i_d_T_ij[idx(ir, iz, Z0)] = 0.0;
+                d_NBIHeatTerm_e_d_n_ij[idx(ir, iz, Z0)] = 0.0;
+                d_NBIHeatTerm_e_d_T_ij[idx(ir, iz, Z0)] = 0.0;
+            }
+        }
+    }
+
+    // Set the beam power, either constant or time-dependent
+    real_t powerNow = 0;
+    if (this->Power_Profile != nullptr) {
```
hoppe93: Why is it handled like this? Why can't the ``Power_Profile`` also be used for a time-constant power?

hoppe93 (changes requested):
Very nicely done! I made some comments which we should discuss at some point.

PR #460, opened 2025-11-20
Newton method backtracking algorithm

### LINKED ISSUE #72: Line search in SolverNonLinear
We should investigate whether we can implement a line search method in the nonlinear solver that could improve stability (and possibly accelerate convergence).  

An issue I foresee is that the different equations have different characteristic scales, so that the residual of for example density or kinetic equations (which have a characteristic scale ~10^20) would be much larger than the residual of, say, a poloidal flux equation (which takes order-unity values). Probably the residuals of different unknowns should be individually normalized somehow.

hoppe93:
This pull request extends the Newton solver with a backtracking algorithm to (theoretically) allow for the acceleration of Newton iteration. Due to the inhomogeneity of the DREAM equation system, the backtracking algorithm may not always work so well though and probably more work would be needed to make it useful.

Before merge, documentation should be written!


src/Solver/Backtracker.cpp
```diff
@@ -0,0 +1,268 @@
...
+			this->monitor[i] = uqh->GetUnknownID(monitor[i]);
+	} else {
+		// If the list is empty, we consider all non-trivial unknowns
+		this->nMonitor = nnu;
+		this->monitor = new len_t[this->nMonitor];
+		for (len_t i = 0; i < this->nMonitor; i++)
+			this->monitor[i] = nu[i];
+	}
+
+    this->f0 = new real_t[nnu];
```
OEmbreus: allocates space for `nnu`; later uses loop with `nMonitor`

OEmbreus (changes requested):
This update is looking really nice, finally! I still remember testing this branch all those years ago. :) Great that you have been putting the work in to push it over the finish line... just unfortunate that it doesn't seem to help stabilize the worst cases.
 
I discovered a few potential issues that should be addressed. The most pressing issue is the comment on the revised REDO loop below.

A section should be added to (probably) `solver.rst` in the docs to describe the `newton_step_adjuster` options.

Also, it would be awesome if we could add a pathological case under `examples/` with convergence issues where these methods are demonstrated.

src/Solver/Backtracker.cpp
```diff
@@ -0,0 +1,268 @@
...
+	if (this->nMonitor > 0) {
+		this->monitor = new len_t[this->nMonitor];
+		for (len_t i = 0; i < this->nMonitor; i++)
+			this->monitor[i] = uqh->GetUnknownID(monitor[i]);
+	} else {
+		// If the list is empty, we consider all non-trivial unknowns
+		this->nMonitor = nnu;
+		this->monitor = new len_t[this->nMonitor];
+		for (len_t i = 0; i < this->nMonitor; i++)
+			this->monitor[i] = nu[i];
```
IdaEkmark, line 51: Just to make sure, this is the same as the following, right?
`this->monitor[i] = uqh->GetUnknownID(nu[i]);`

src/Solver/PhysicalStepAdjuster.cpp
```diff
@@ -0,0 +1,173 @@
...
+	for (len_t it=0; it<N; it++) {
+		const len_t id = nontrivial_unknowns[it];
+		FVM::UnknownQuantity *uq = unknowns->GetUnknown(id);
+		len_t NCells = uq->NumberOfElements();
+
+		// check whether unknown it is a non-negative quantity
+		bool isNonNegativeQuantity = false;
+		for (len_t it_nn = 0; it_nn < N_nn; it_nn++)
+			if(id==ids_nonNegativeQuantities[it_nn])
+				isNonNegativeQuantity = true;
```
IdaEkmark, line 122: Don't we want to break the loop after `isNonNegativeQuantity = true;`?

src/Solver/PhysicalStepAdjuster.cpp
```diff
@@ -0,0 +1,173 @@
...
+				// require x1 > threshold*x0
+				real_t maxStepAtI = MaximalStepLengthAtGridPoint(x0[offset+i], dx[offset+i], threshold);
+				// if this is a stronger constaint than current maxlength, override
+				if(maxStepAtI < maxStepLength && maxStepAtI>0){
+					maxStepLength = maxStepAtI;
+					this->limitingUnknown = id;
+				}
+			}
+
+		if(nonNegativeZeff && id==id_ni){
```
IdaEkmark, line 137: Is `nonNegativeZeff` ever `false`? It is set to `true` at the beginning of the method and I don't see where it would change otherwise

src/Solver/PhysicalStepAdjuster.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+	// add those quantities which we expect to be non-negative
+	// T_cold and n_cold will crash the simulation if negative, so they should always be added
+	bool nonNegativeZeff = true;
+
+	const len_t N = nontrivial_unknowns.size();
+	const len_t N_nn = ids_nonNegativeQuantities.size();
+	len_t offset = 0;
+	// sum over non-trivial unknowns
+	for (len_t it=0; it<N; it++) {
```
IdaEkmark, line 113: Would it be more efficient to have an array with the `non_trivial_unknowns` that should not be negative and have this be a loop over `N_nn` instead of `N` and you can skip the inner for loop on line 120?

Similarly, you could have a variable for the index of id_ni, so that that part of this for loop is still intact

src/Solver/PhysicalStepAdjuster.cpp
```diff
@@ -0,0 +1,173 @@
...
+				isNonNegativeQuantity = true;
+
+		// Quantities which physically cannot be negative, require that they cannot
+		// be reduced by more than some threshold in each iteration.
+		if(isNonNegativeQuantity)
+			for(len_t i=0; i<NCells; i++){
+				// require x1 > threshold*x0
+				real_t maxStepAtI = MaximalStepLengthAtGridPoint(x0[offset+i], dx[offset+i], threshold);
+				// if this is a stronger constaint than current maxlength, override
+				if(maxStepAtI < maxStepLength && maxStepAtI>0){
```
IdaEkmark, line 131: What does `maxStepAtI>0` do? Doesn't this only happen if `threshold>1`?

src/Solver/PhysicalStepAdjuster.cpp
IdaEkmark: I see now the problem is `offset`, you can ignore this comment

src/Solver/PhysicalStepAdjuster.cpp
```diff
@@ -0,0 +1,173 @@
...
+ */
+PhysicalStepAdjuster::~PhysicalStepAdjuster() { }
+
+
+/**
+ * Check if the Newton step needs to be updated.
+ */
+bool PhysicalStepAdjuster::AdjustmentNeeded(
+	const len_t, Vec&, FVM::BlockMatrix*
+) {
```
IdaEkmark, line 45: What is the purpose of this function?

doc/notes/discretisation.tex
```diff
@@ -650,15 +682,15 @@
...
-    given in {\em Numerical Recipes} and is based on the optimization of an
-    equivalent one-dimensional problem. The purpose of the method is to
-    determine the optimal damping factor $\lambda$ such that the Newton update
+    the Newton solver may overshoot and first oscillate around the solution,
+    or even diverge completely. One method of accelerating convergence is given
+    in {\em Numerical Recipes} and is based on the optimization of an equivalent
+    one-dimensional problem. The purpose of the method is to determine the
+	optimal damping factor $\lambda$ such that the Newton update
     \begin{equation}
         \bb{x}^{(k+1)} = \bb{x}^{(k)} + \lambda\delta\bb{x},
```
IdaEkmark, line 691: Should it not be ${x}^{(k+1)} = {x}^{(k)} - \lambda\delta{x},$? For the maximal step length, you check if $\delta x$ is positive to make sure $x_1$ would not be negative

src/Solver/Backtracker.cpp
```diff
@@ -0,0 +1,268 @@
...
+ * detailed results are available in 'dec'.
+ *
+ * dec: Each element corresponds to a non-trivial unknown of the
+ *      system. If the element is 'true', the Newton step is leading
+ *      to a more optimal value for the unknown.
+ */
+bool Backtracker::IsDecreasing(bool *dec) {
+    bool v = true;
+    for (len_t i = 0; i < this->nMonitor; i++) {
+        dec[i] = (f1[i] <= f0[i]-ALPHA*gradf_deltax[i]);
```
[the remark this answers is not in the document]
IdaEkmark: From my understanding of the code, $\delta x_{\rm code}=-\delta x_{\rm doc}$. I found it looking at the `MaximalStepLengthAtGridPoint`, but I might have misunderstood

src/Solver/Backtracker.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+/**
+ * Determine if the Newton step needs to be adjusted.
+ */
+bool Backtracker::AdjustmentNeeded(
+	const len_t iteration, Vec &F, FVM::BlockMatrix *jac
+) {
+	// If the physical step adjuster was previously enabled, we have
+	// now taken that step and can proceed to the next Newton iteration.
+	if (this->maximalPhysicalStepLength < 1)
```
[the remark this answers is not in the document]
IdaEkmark: Doesn't `(this->maximalPhysicalStepLength < 1` explicitly mean an adjustment is needed? Since then `lambda=maximalPhysicalStepLength`

src/Solver/Backtracker.cpp
```diff
@@ -0,0 +1,268 @@
...
+            real_t g1  = f1[i];
+            real_t g2  = f2[i];
+            real_t l1  = lambda1, l2 = lambda2;
+
+            real_t l1_2 = l1*l1, l1_3 = l1_2*l1;
+            real_t l2_2 = l2*l2, l2_3 = l2_2*l2;
+
+            real_t den = l1_2*l2_2*(l1-l2);
+            real_t a = (l2_3*g1 - l1_3*g2 + l1*l2*(l1-l2)*gp0+ (l1_2-l2_2)*g0) / den;
+            real_t b = (l1_3*g2 - l2_3*g1 + l1*l2*(l2_3-l1_3)*gp0 + (l2_3 - l1_3)*g0) / den;
```
IdaEkmark, line 168: The numerators are different from the documentation. The numerators here are
a: $\lambda_2^3 g(\lambda_1)-\lambda_1^3 g(\lambda_2) + \lambda_1\lambda_2(\lambda_1-\lambda_2)g'(0)+(\lambda_1^2-\lambda_2^2)g(0)$
b: $\lambda_1^3 g(\lambda_2)-\lambda_2^3 g(\lambda_1) + \lambda_1\lambda_2(\lambda_2^3-\lambda_1^3)g'(0)+(\lambda_2^2-\lambda_1^2)g(0)$

For a, the first two terms are different, for b the third term is different

src/Solver/Backtracker.cpp
```diff
@@ -0,0 +1,268 @@
...
+		else if (lambda > 0.5*lambda1)
+			lambda = 0.5*lambda1;
+	}
+    
+    return lambda;
+}
+
+/**
+ * Evaluate the target function f = -F*F, where F is
+ * the residual function.
```
IdaEkmark, line 197: In the documentation, the target function is f = +F*F

src/Solver/Backtracker.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+        // On the very first backtracking iteration, also calculate
+        // the gradient of f, dotted with the Newton step 'delta-x',
+        // and store g(0) = f
+        if (nIteration == 0) {
+            gradf_deltax[i] = -f1[i];
+            f0[i] = 0.5*f1[i];
+		}
+
+        f1[i] *= 0.5;
```
IdaEkmark, line 224: Why multiply by 0.5? 

Also, to save space, line 224 could be moved before the if statement on line 219, and line 221 could be removed, right? Not very important though

IdaEkmark (commented):
It looks good, but I agree with Ola and have some comments/questions of my own

PR #464, opened 2025-12-05
Suppport for multiple magnetic field minima/maxima in fluid mode

hoppe93:
DREAM has for a long time only allowed magnetic fields to have one maximum/minimum along each field line. This is motivated by the severe difficulties associated with bounce averaging quantities in such magnetic fields. Flux surface averages are however much more straightforward to do in such fields, and so it should still be possible to run fluid simulations with such magnetic fields.

This pull request relaxes the restriction on magnetic field optima to only apply when a momentum grid is also used in the simulation. Pure fluid simulations are thus exempt from the restriction. The code will now explicitly look for global minima/maxima in the magnetic field and use them for fluid calculations. The most crucial quantities which require information about the magnetic field extrema in fluid mode are the trapped-passing boundary (``xi0T``) and the effective passing fraction, both of which generalize easily to only be evaluated with the global maximum of the magnetic field strength.


lvotta98:
When trying to use it with ITER scenarios `H26` and `DTHMode24` it raise an error `fsolver.c:116: x_minimum must lie inside interval (lower < x < upper) (GSL_EINVAL)`

OEmbreus:
> > > I feel a bit uneasy about there being no unit test for the `NumericBRadialGridGenerator`. For example, just skimming the code to figure out what it does, I immediately noticed an error in `NumericBRadialGridGenerator::EvaluateGeometricQuantities` where `NablaR2` is evaluated with a `prefactor R*R` instead of `ROverR0*ROverR0` (since the jacobian is normalized to `R0` -- compare with `EvaluateNablaR2`).
> > > Wouldn't it be reassuring to generate high-resolution magnetic-field data representing the parametrized field used in `AnalyticBRadialGridGenerator` and verify that NumericB evaluates the same `GeometricQuantities` (within the expected precision)?
> > > Doesn't have to be in this PR, which I think looks good now overall :) perhaps sneak in the geometricquantities bug fix?
> > 
> > 
> > Isn't what you described essentially just the `numericmag` test? (`tests/physics/numericmag`)? It uses a high-resolution numerical version of the analytical field and achieves agreement to within 0.002%.
> > Regarding the place in the code which you refer to, I cannot say for certain that it is correct, but considering the result of the test and the way in which the code is written, I would suspect that the desired factor `R0` appears in one of the other quantities?
> 
> Sorry, I read your comment more carefully now, and indeed, a C++ test of the same could be useful indeed :P We'll see if someone can find some time in the future...

Oh, it was mostly me being silly! I had a look around and didn't find the test at first (among the cxx tests). Looks like we are testing exactly what I was looking for :) 
I made an edit above. There's 2 apparent bugs in the code, but perhaps it works out correctly in the end somehow. 1e-5 error sounds reassuring indeed!

OEmbreus:
> > I feel a bit uneasy about there being no unit test for the `NumericBRadialGridGenerator`. For example, just skimming the code to figure out what it does, I immediately noticed an error in `NumericBRadialGridGenerator::EvaluateGeometricQuantities` where `NablaR2` is evaluated with a `prefactor R*R` instead of `ROverR0*ROverR0` (since the jacobian is normalized to `R0` -- compare with `EvaluateNablaR2`).
> > Wouldn't it be reassuring to generate high-resolution magnetic-field data representing the parametrized field used in `AnalyticBRadialGridGenerator` and verify that NumericB evaluates the same `GeometricQuantities` (within the expected precision)?
> > Doesn't have to be in this PR, which I think looks good now overall :) perhaps sneak in the geometricquantities bug fix?
> 
> Isn't what you described essentially just the `numericmag` test? (`tests/physics/numericmag`)? It uses a high-resolution numerical version of the analytical field and achieves agreement to within 0.002%.
> 
> Regarding the place in the code which you refer to, I cannot say for certain that it is correct, but considering the result of the test and the way in which the code is written, I would suspect that the desired factor `R0` appears in one of the other quantities?

Still, it would be nice to unit test the geometric calculations in the kernel. A conductivity test in a very modestly shaped magnetic field is relatively blunt. 

I see that `numericmag` generates a LUKE format magnetic data file. Could we save a reference field somewhere (`tests/static` or something?) and use in a cxx test?

PR #467, opened 2025-12-10
Fix bug in plotting of radial and time profiles in FluidQuantity

hoppe93:
This pull request fixes a bug which caused the data of fluid quantities to accumulate on subsequent calls to ``plotRadialProfile()`` and ``plotTimeProfile()``, giving incorrect plots the second and later times that those routines are called.


ekmanlinn (approved):
Looks good!

PR #471, opened 2026-01-12
Implement adaptive Rechester–Rosenbluth heat transport operator

Shengyang-HUST:
This PR implements an adaptive Rechester–Rosenbluth heat transport operator with an MHD-like transition.

Main changes:
- Add `HeatTransportRRAdaptiveMHDLike` operator (hpp/cpp)
- Extend RR heat transport interfaces and implementation
- Wire the new transport term through `Transport.cpp`

### DIFF
--- a/examples/transport-rr/generate_consistent.py
+++ b/examples/transport-rr/generate_consistent.py
@@ -13,6 +13,7 @@
 # ###################################################################
 
 import numpy as np
+import math
 import sys
 
 sys.path.append('../../py/')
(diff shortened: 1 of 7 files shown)


examples/transport-rr/generate_consistent.py
hoppe93: It would be better if you create a completely example for this operator, in a new directory. You can copy the script from ``transport-rr``, but it would be good to keep the original script.

PR #478, opened 2026-01-16
Bootstrap current implementation

peterhalldestam:
## Summary

This pull request implements the **bootstrap current model by [Redl et al. (2021)](https://doi.org/10.1063/5.0012664)** in DREAM.
The model provides an analytical expression for the flux surface averaged bootstrap current, fully consistent with the neoclassical conductivity formulation from the same reference, which is already available in DREAM.

## Details

* Added calculation of the bootstrap current density, `j_bs`, following *Redl et al. (2021)*.
* Allows optional inclusion of the bootstrap current density in the total current density.
* Initialise either from provided total current density or Ohmic current density.
* Verified against implementation in IDA using data from ASDEX Upgrade.

### DIFF
--- a/include/DREAM/Equations/SPIHandler.hpp
+++ b/include/DREAM/Equations/SPIHandler.hpp
@@ -97,6 +97,7 @@ namespace DREAM{
         int_t *nbrShiftGridCell=nullptr;
         int_t *nbrShiftGridCellPrescribed=nullptr;
         real_t *shift_store=nullptr;
+        real_t *shift_store_major_radius=nullptr;
         real_t *ncoldPrevious=nullptr;
         real_t *TcoldPrevious=nullptr;
         real_t *YpdotPrevious=nullptr;
@@ -187,6 +188,9 @@ namespace DREAM{
 
         real_t* GetDrift()
         { return this->shift_store; }
+
+        real_t* GetDriftMajorRadius()
+        { return this->shift_store_major_radius; }
 
         void Rebuild(real_t dt, real_t t);
 
--- a/py/DREAM/Output/FluidQuantity.py
+++ b/py/DREAM/Output/FluidQuantity.py
@@ -226,52 +226,92 @@ class FluidQuantity(UnknownQuantity):
             raise OutputException("Cannot plot a scalar value. r = {}, t = {}.".format(r, t))
 
 
-    def plotPoloidal(self, ax=None, show=None, t=-1, colorbar=True, return_contours=False, displayGrid=False, maxMinScale=True, logscale=False, **kwargs):
+    def plotPoloidal(self, ax=None, show=None, t=-1, colorbar=True, return_contours=False, displayGrid=False, maxis = False, shifted = True, maxMinScale=True, levels = None, logscale=False, **kwargs):
         """
-        Plot the radial profile of this quantity revolved over a 
-        poloidal cross section at the specified time step. 
-        NOTE: Currently assumes a cylindrical flux surface geometry!
-        
+        Plot the radial profile of this quantity revolved over a
+        poloidal cross section at the specified time step.
+
         :param matplotlib.pyplot.Axis ax:   Matplotlib axes object to use for plotting.
         :param bool show: If 'True', shows the plot immediately via a call to 'matplotlib.pyplot.show()' with 'block=False'. If 'None', this is interpreted as 'True' if 'ax' is also 'None'.
         :param int t: Time index to plot.
         :param matplotlib.pyplot.Colorbar colorbar: Specify wether or not to include a colorbar.
-        :param bool displayGrid: Specify wether or not to display a polar grid in the plot.
+        :param bool return_contours: Specify wether or not to return the contours
+        :param bool displayGrid: Specify wether or not to display the flux surfaces separating the grid cells in the plot.
+        :param bool maxis: Specify wether or not to display the magnetic axis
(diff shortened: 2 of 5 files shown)


doc/notes/bootstrap_current.tex
```diff
@@ -0,0 +1,121 @@
...
+    In \DREAM, one may calculate safety factor using the already implemented method \texttt{RadialGrid::SafetyFactorNormalized}, which uses a formula (see doc/notes/theory.tex under the ``Safety factor'' section) that explicitly relates $q$ to be inversily proportional to the total current $I(r)$ enclosed by a flux surface labeled $r$.
+    However, when instead using the approximation (found in page 167 of Wesson's ``Tokamaks 2nd ed.''), namely
+    \begin{align}
+        \label{eq:safety-factor-wesson}
+        q(r)\approx\varepsilon\frac{B}{B_\theta}=\varepsilon\sqrt{1+\frac{4\pi^2G^2}{\psi'^2}},
+    \end{align}
+    where $\varepsilon= r/R$ is the inverse aspect ratio, a much better agreement with the implemented formula is observed when cross-checking with reconstructions of the bootstrap current in AUG discharges (AUGD/IDF which uses the Redl-Sauter model).
+    \textcolor{red}{Whether we want to use one or the other formulas for the safety factor is to be determined, but for the moment, we use Eq.~\eqref{eq:safety-factor-wesson}.}
```
hoppe93: The formula from Wesson is valid in the large-aspect ratio limit, while the DREAM expression for q should be valid at arbitrary radii. The two should agree well near the magnetic axis. It feels somewhat awkward to use the Wesson formula when the DREAM formula should be more correct, and it is strange if the Wesson formula agrees better with equilibrium reconstructions. Could it perhaps be that the safety factor is not reconstructed directly, but is computed using the Wesson formula from reconstructions of e.g. the magnetic field?

py/DREAM/Settings/Equations/BootstrapCurrent.py
```diff
@@ -0,0 +1,80 @@
...
+        self.setInitMode(initMode)
+
+
+    def setMode(self, mode, initMode=None):
+        """
+        Specifies whether to include the bootstrap current as a contribution to the total
+        current density. If enabled, this contribution is calculated using the Redl-Sauter
+        model, which is based on A. Redl et al (DOI: https://doi.org/10.1063/5.0012664).
+        """
+        if mode in [BOOTSTRAP_MODE_DISABLED, BOOTSTRAP_MODE_ENABLED]:
```
hoppe93: Perhaps one could also allow ``mode = True`` and associate it with ``BOOTSTRAP_MODE_ENABLED`` (and perhaps same for ``False``)?

py/DREAM/Settings/Equations/BootstrapCurrent.py
```diff
@@ -0,0 +1,80 @@
+# Settings for the Bootstrap current
+from . EquationException import EquationException
+from . UnknownQuantity import UnknownQuantity
+
+from . PrescribedInitialParameter import PrescribedInitialParameter
+
+BOOTSTRAP_MODE_DISABLED = 1
+BOOTSTRAP_MODE_ENABLED = 2
```
hoppe93: Would a better name for this option be ``BOOTSTRAP_MODE_REDL`` instead? In case someone would like to implement another bootstrap model in the future?

hoppe93 (changes requested):
Very nice work! :D I have a few comments and questions which we can discuss before merging.

examples/bootstrap/ASDEXU.py
```diff
@@ -0,0 +1,128 @@
...
+    :param int nr: Radial grid resolution to use if ``r = None``.
+    """
+    global a
+
+    if r is None:
+        r = np.linspace(0, a, nr)
+
+    j0 = 1
+
+    # return r, j0 * (1 - (1-0.001**(1/0.41))*(r/a)**2)**0.41
```
IdaEkmark: Here we could also use some clean up

src/Atomics/nistdata_ionization.cpp
IdaEkmark: As previous comment

src/Equations/Fluid/BootstrapIonDensityTerm.cpp
```diff
@@ -0,0 +1,78 @@
...
+    } else {
+        dCoefficient *= bs->p[ir] / bs->n[ir] - ( 1. + alpha ) * bs->Wi[rOffset + ir];
+        dCoefficient -= l31 * 2./3. * dalpha * bs->Wi[rOffset + ir] / bs->Ni[rOffset + ir];
+        if (derivId == id_ncold)
+            dCoefficient += l31 * ( bs->Tcold[ir] - bs->p[ir] / bs->n[ir] ) / bs->n[ir];
+        else if (derivId == id_Tcold)
+            dCoefficient += l31 * bs->ncold[ir] / bs->n[ir];
+        else if (derivId == id_Ni)
+            dCoefficient += l31 * ( (1. + alpha ) * bs->Wi[rOffset + ir] / ( 1.5 * bs->Ni[rOffset + ir] * bs-> Ni[rOffset + ir]) - bs->p[ir] / (bs->n[ir] * bs->n[ir]) );
+        else if (derivId == id_Wi) // IE: Added: l31 / (1.5 * bs->n[ir] * Constants::ec) -
```
IdaEkmark: Remove comment

IdaEkmark (commented):
I mostly have comments about clean-up. In general the implementation looks great!

PR #479, opened 2026-01-16
Orbit-averaged general knock-on collisions: theory notes, utilities, and unit tests

### LINKED ISSUE #470: Implement general knock-on source
Isn't it about time we implement proper knock-ons? 
- full bounce-averaged term accounting for runaway energy and pitch distribution
- external runaways should contribute with the already-implemented RP source term
- it should probably be treated explicitly in the time integration scheme, since its kinetic jacobian block is described by a full matrix.

The theory is worked out, it's just a matter of implementing and testing.
There are some question marks with regards to implementation:
The bounce averages appearing in the Rosenbluth-Putvinski term are currently calculated in the FluxSurfaceAverager as part of the FVM library. This feels a bit malplaced, and should perhaps find a new home in the DREAM library? 
There's quite a few calculations to be implemented for the new equation term, and storing the results may be heavy on memory. I think we should ensure that we don't allocate memory for these unless it's going to be used (unlike how the current term is treated?).

OEmbreus:
**Summary**

This PR introduces the core physics and numerical infrastructure for an orbit-averaged formulation of the general knock-on collision kernel. It consists of three parts:

- theory notes deriving the orbit-averaged delta function in detail
- a new `KnockOnUtilities` module implementing the derived expressions
- a unit test module validating the implementation against analytic properties and limiting cases

The actual equation term is not included here and will follow in a separate PR.

**Physics included**
The utilities implement the bounce-, gyro- and FVM-cell-averaged kinematic delta function appearing in the electron-electron scattering problem. Accurately evaluating this angular dependence after averaging turns out to be the main technical difficulty in implementing knock-on collision terms.

With this contribution in place, implementing actual collision operators (general, Chiu-Harvey, Rosenbluth-Putvinski) should be comparatively straightforward.

The code notation and documentation follow the conventions used in the theory notes, with the intent that the implementation can be read side-by-side with the derivations.

**Testing**
The test module checks both physics invariants (symmetry and conservation properties) and numerical aspects (convergence with respect to resolution parameters). Tests cover general magnetic geometry as well as the cylindrical (constant-B) limit, and include benchmarks against the existing Rosenbluth–Putvinski implementation.


fvm/Grid/FluxSurfaceAverager.avalancheDelta.cpp
```diff
@@ -54,18 +54,7 @@ real_t hIntegrand(real_t theta, void *par){
...
-
-    real_t g = params->gamma;
-    real_t xi = RESign*sqrt((g-1)/(g+1));
-    real_t xi0 = xi0Star(BOverBmin,g,RESign);
-    real_t xiOverXi0 = xi/xi0;    
-    real_t sqrtgOverP2 = MomentumGrid::evaluatePXiMetricOverP2(xiOverXi0,BOverBmin);
-
-    // 2*pi for the trivial phi integral
-    return 2*M_PI * xiOverXi0 * Jacobian * sqrtgOverP2 / (dxi * Vp);
+    return 2 * M_PI * BOverBmin * Jacobian;
```
IdaEkmark, line 57: If I understand correctly, you have used 
$$\sqrt{g}=p^2\frac{B}{B_{\rm min}}\frac{\xi_0}\xi\mathcal{J}\Rightarrow\frac\xi{\xi_0}=\frac{p^2}{\sqrt{g}}\frac{B}{B_{\rm min}}\mathcal{J}$$
 to make this expression simpler. Shouldn't the Jacobian be squared then? 

Also, I note that you have moved 1 / (dxi * Vp) to line 210, but there you have also added a factor $2\pi$ without removing it here. Should this factor be in both places?

tests/cxx/tests/DREAM/RunawayFluid.cpp
```diff
@@ -509,7 +509,7 @@ bool RunawayFluid::CompareConnorHastieRateWithTabulated() {
     DREAM::FVM::UnknownQuantityHandler *uqn = REFluid->GetUnknowns();
     len_t id_n_cold = uqn->GetUnknownID(DREAM::OptionConstants::UQTY_N_COLD);
     real_t ncold = uqn->GetUnknownData(id_n_cold)[0];
-    this->PrintStatus("ncold = %e", ncold);
+    // this->PrintStatus("ncold = %e", ncold);
     //real_t ncold = 4.6009999999999997e+21;
```
IdaEkmark, line 513: Can these comments be removed?

include/DREAM/Equations/KnockOnUtilities.hpp
```diff
@@ -0,0 +1,211 @@
...
+
+/**
+ * Cosine of the angle between incident electron and knock-on,
+ * as function of their respective momenta.
+ *   p1: momentum magnitude of the incident electron (assumed to
+ *       collide with stationary electron).
+ *   p: knock-on momentum magnitude.
+ */
+inline real_t EvaluateXiStar(real_t p, real_t p1) {
+    // handle the well-behaved (and common) limit of infinitely energetic incident electron
```
IdaEkmark, line 43: Should this be one line lower? `g` is used also for incident electrons with finite energy

include/DREAM/Equations/KnockOnUtilities.hpp
```diff
@@ -0,0 +1,211 @@
...
+#include "DREAM/Constants.hpp"
+#include "FVM/Grid/Grid.hpp"
+#include "FVM/Grid/RadialGrid.hpp"
+
+namespace DREAM::KnockOnUtilities {
+
+// ============================================================================
+// Low-level analytic building blocks (no grid dependence)
+// ============================================================================
+namespace Kinematics {
```
IdaEkmark, line 32: Could the methods in Kinematics be moved to the cpp-file? My personal preference is to only have very simple method implementation (consisting of only one line) in hpp-files, and have methods with more than one line in the cpp file. Here, `EvaluateXiStar`, `EvaluateMollerDifferentialCrossSection`, `EvaluateMollerFlux` and `MaximumKnockOnMomentum` take up fairly much space for a hpp file

src/Equations/KnockOnUtilities.cpp
```diff
@@ -0,0 +1,829 @@
...
+real_t KnockOnUtilities::AnalyticDelta::EvaluateXiIntegral(
+    real_t t1, real_t t2, real_t xi1, real_t xi_star
+) {
+    real_t a = xi_star * xi1;
+    real_t b2 = (1 - xi_star * xi_star) * (1 - xi1 * xi1);
+
+    // handle singular (but analytically well-posed) case of
+    // EvaluateXiAntiderivative when the denominator is 0
+    if (b2 <= RELAXED_EPS) {
+        real_t a = xi_star * xi1;
```
IdaEkmark, line 74: This is already defined, right?

src/Equations/KnockOnUtilities.cpp
```diff
@@ -0,0 +1,829 @@
...
+ * Analytic pitch integration of the gyro-averaged kinematic delta function,
+ * from xi=t1 to xi=t2 (clipped to the kinematically available region [a-b, a+b],
+ * in which case the values of F are limited to +/- 0.5).
+ * Evaluates the variable "D" defined in the theory notes.
+ */
+real_t KnockOnUtilities::AnalyticDelta::EvaluateXiIntegral(
+    real_t t1, real_t t2, real_t xi1, real_t xi_star
+) {
+    real_t a = xi_star * xi1;
+    real_t b2 = (1 - xi_star * xi_star) * (1 - xi1 * xi1);
```
IdaEkmark, line 69: Minor thing and not at all necessary to change, but could be nice to have the factors in the same order as in EvaluateXiAntiderivative

src/Equations/KnockOnUtilities.cpp
```diff
@@ -0,0 +1,829 @@
...
+        z2 = t1 * xi_star + sqrt(1 - t1 * t1) * sqrt(1 - xi_star * xi_star);
+    }
+}
+
+/**
+ * Evaluates the analytically gyro- and xi-cell-averaged kinematic
+ * delta function at a fixed poloidal location, characterized
+ * by its BOverBmin and xi0Cutoff values.
+ *
+ * In the theory notes, the returned value corresponds to the D_j function.
```
IdaEkmark, line 145: I thought `EvaluateXiIntegral` returned the $D_j$ function value, since it returns $F_2-F_1$, same as in equation (84) in the theory notes. Here, you divide `EvaluateXiIntegral` with `(xi0_f2 - xi0_f1)`, unlike equation (84). Am I missing something?

src/Equations/KnockOnUtilities.cpp
```diff
@@ -0,0 +1,829 @@
...
+    // employ an exacter constraint by inserting values explicitly.
+    real_t t1;
+    real_t t2;
+    AnalyticDelta::ComputeXiIntegrationBounds(t1, t2, xi0_f1, xi0_f2, BOverBmin, xi0Cutoff);
+
+    // handle the xi01 == 1 case (used in the approximate collision terms--RP and CH)
+    if (fabs(fabs(xi01) - 1.0) < RELAXED_EPS) {
+        xi1_over_xi01 = 1.0;
+        // exact solution of z2==1 (since z = xi1 = 1)
+        if (xi01 > 0.0 && !(t1 < xi_star && xi_star < t2)) {
```
IdaEkmark, line 172: Isn't $z_2=1$ when $t_1<\xi^\star<t_2$? Is the `!` wrong or is the comment above wrong or am I missing something?

src/Equations/KnockOnUtilities.cpp
```diff
@@ -0,0 +1,829 @@
...
+            return 0.0;
+        }
+        // exact solution of z1==-1 (since z = xi1 = -1)
+        if (xi01 < 0.0 && !(t1 < -xi_star && -xi_star < t2)) {
```
IdaEkmark, line 176: Isn't $z_1=1$ when $t_1<-\xi^\star<t_2$? Is the `!` wrong or is the comment above wrong or am I missing something?

src/Equations/KnockOnUtilities.cpp
IdaEkmark: Should this comment say 
`The returned value corresponds to the D_j function in the theory notes over Delta xi`?

src/Equations/KnockOnUtilities.cpp
```diff
@@ -0,0 +1,829 @@
...
+
+/**
+ * Evaluates the orbit-averaged Delta including trapping corrections on
+ * the provided grid.
+ *
+ * Uses xi01 on grid centers at index l and xi0_{j-1/2}, xi0_{j+1/2} at index j.
+ * This function is a grid-aware convenience wrapper around
+ * EvaluateOrbitAveragedDeltaWithTrappingCorrection().
+ *
+ * Corresponds to the matrix element Delta_{j l} in the theory notes.
```
IdaEkmark, line 563: I find no $\Delta_{j l}$ in the theory notes

tests/cxx/tests/DREAM/KnockOn.cpp
```diff
@@ -0,0 +1,812 @@
...
+                // len_t j = xi0_indices[idx_j];
+                for (len_t idx_l = 0; idx_l < 6; idx_l++) {
+                    len_t l = xi01_indices[idx_l];
+                    real_t delta_default =
+                        DREAM::KnockOnUtilities::EvaluateDeltaMatrixElementOnGrid(
+                            ir, xi_star, j, l, grid, grid, nPointsIntegral,
+                            DREAM::KnockOnUtilities::MIDPOINT_RULE
+                        );
+                    real_t delta_hires = DREAM::KnockOnUtilities::EvaluateDeltaMatrixElementOnGrid(
+                        ir, xi_star, j, l, grid, grid, 300, DREAM::KnockOnUtilities::MIDPOINT_RULE
```
IdaEkmark, line 286: Is there  a reason why 300 is hard coded while `nPointsIntegral` is given a variable?

IdaEkmark (changes requested):
Looks good, but I have some questions.

PR #480, opened 2026-01-18
RadialGridGenerator unit tests and NumericB bug fixes

OEmbreus:
Add unit testing of the `AnalyticBRadialGridGenerator` and `NumericBRadialGridGenerator`.
Adds a LUKE equilibrium file for unit testing, constructed to be identical to the AnalyticB field that is already used in various tests.

A minor bug was discovered in the `NumericBRadialGridGenerator`, and resolved. 

Fixes to the `numericmag.constructMagneticField` python routine for saving parametric magnetic fields to LUKE format, which had some minor issues.


tests/cxx/tests/FVM/RadialGridGenerator.cpp
```diff
@@ -0,0 +1,204 @@
...
+ * with different radial grid generators, such as Cylindrical, AnalyticB and Numerical.
+ */
+#include "RadialGridGenerator.hpp"
+#include "FVM/Grid/NumericBRadialGridGenerator.hpp"
+#include "UnitTest.hpp"
+
+using namespace DREAMTESTS::FVM;
+
+namespace {
+bool TestGeometricQuantitiesAtTheta(
```
hoppe93: Why is this function placed in its own namespace, and not as part of the ``RadialGridGenerator`` class? If it was a member of ``RadialGridGenerator`` it would have access to the ``PrintError()`` and ``PrintOK()`` methods.

OEmbreus:
> Thanks a lot for implementing this! :D It runs great on my machine!
> 
> I added two comments/questions. They don't necessarily require changes.
> 
> Another general question is whether it is perhaps better to hold off with merging this until the `multiple-Bminmax` branch has been merged to `master`? Or at least until it has been fully reviewed? Doing so would probably make it easier for the reviewers to get an overview of the changes.

Yes, let’s hold off on merging this. We can wait for minmax and then merge to main after.
Thanks for the comments, good stuff. Will fix tonight!

tests/cxx/tests/FVM/RadialGridGenerator.cpp
```diff
@@ -0,0 +1,204 @@
...
+    // A relatively loose tolerance is used, so that we can
+    // get away with using a smaller dataset. Higher-resolution
+    // data yields smaller errors.
+    real_t test_tol = 1e-4;
+
+    len_t nr = 4;
+    real_t ntheta_interp = 200;
+
+    // the DeltaPrime value of the example grid
+    real_t DeltaPrime = 0.3;
```
OEmbreus: I didn’t like having to hardcode this value, which is also used locally in Unittests.grids.cpp as a constexpr (and ofc then also in the python script to generate the geometry…). Not sure how to solve it in an elegant way, perhaps as a protected getter in UnitTest.hpp? ’real_t test_grid_max_shafranov_shift=0.3;’? 
no biggie i guess, but still. @hoppe93

tests/cxx/tests/FVM/RadialGridGenerator.cpp
```diff
@@ -1,16 +1,61 @@
...
 using namespace DREAMTESTS::FVM;
 
-namespace {
-bool TestGeometricQuantitiesAtTheta(
-    len_t ir, real_t theta, DREAM::FVM::RadialGrid *rg, real_t test_tol
+/**
+ * Compare two floating-point values with an atol and rtol using the same
+ * conventions as, for example, in numpy and scipy.
+ */
+static bool nearlyEqual(real_t a, real_t b, real_t atol, real_t rtol) {
```
OEmbreus: you could almost imagine the base class, UnitTest, to provide a utility like this (as well as printComparison below), to enable uniform comparison methodology and "debug" message styles across the test suite. 

Otherwise we could introduce a UnitTestUtilities with some methods like this (but for the printing, we would need to pass `this` to enable PrintError...)?

Issue #487, opened 2026-01-28
Change implementation of hyperresistive diffusion term


hoppe93:
As noted recently by @peterhalldestam, the current implementation of the hyperresistive diffusion term is incorrect and affects the electric field in an incorrect way. Currently, the hyperresistive diffusion term is added to Faraday's law in the manner

$$
\frac{\partial\psi_ {\rm p}}{\partial t} = -V_{\rm loop} + \frac{\partial}{\partial\psi_{\rm t}}\psi_{\rm t}\Lambda\frac{\partial}{\partial\psi_{\rm t}}\frac{j_\parallel}{B},
$$

with the loop voltage defined according to

$$
V_{\rm loop} = 2\pi\frac{\left\langle\boldsymbol{E}\cdot\boldsymbol{B}\right\rangle}{\left\langle\nabla\phi\cdot\boldsymbol{B}\right\rangle}.
$$

This $V_{\rm loop}$ should however, by definition, directly equal the time-derivative of the poloidal flux, meaning that the loop voltage used in DREAM is inconsistent with the variation of the poloidal flux when hyperresistive diffusion is accounted for. The more correct way of formulating the equations should instead be by adding the hyperresistive diffusion term to Ohm's law

$$
\frac{\left\langle\boldsymbol{E}\cdot\boldsymbol{B}\right\rangle}{\sqrt{\left\langle B^2\right\rangle}} =
\frac{1}{\sigma}\frac{j_\Omega}{B}\sqrt{\left\langle B^2 \right\rangle} -
\frac{\left\langle\nabla\phi\cdot\boldsymbol{B}\right\rangle}{2\pi\sqrt{\left\langle B^2\right\rangle}}\frac{\partial}{\partial\psi_{\rm t}}\psi_{\rm t}\Lambda\frac{\partial}{\partial\psi_{\rm t}}\frac{j_\parallel}{B},
$$

yielding for Farady's law

$$
\frac{\partial\psi_{\rm p}}{\partial t} = -V_{\rm loop}.
$$

OEmbreus:
Scratch the above, that would probably never be feasible/stable. 

Another option:

add a new unknown quantity "the hyperresistive electric field"
```math
\langle \boldsymbol{E}\cdot \boldsymbol{B} \rangle_{hyp} 
= 
[...]

PR #489, opened 2026-01-29
Implement a new form for the hyperresistive diffusion model

hoppe93:
As described in #487, the current implementation of the hyperresistive diffusion model in DREAM appears to be incorrect. This pull request tries to remedy this by moving the hyperresistive term from Faraday's law to Ohm's law. As a consequence, kinetic simulations are no longer compatible with simulations including hyperresistive diffusion.

I have had to reformulate Ohm's law on the form

$$
\frac{\left\langle\boldsymbol{E}\cdot\boldsymbol{B}\right\rangle}
{\sqrt{\left\langle B^2 \right\rangle}}\sqrt{\left\langle B^2 \right\rangle} =
	\frac{1}{\sigma}\frac{j_\Omega}{B}
	\left\langle B^2 \right\rangle
	-
	\frac{1}{V'}
	\frac{\partial}{\partial r}V'\left(
		\frac{\psi_{\rm t}}
		{V'\partial\psi_{\rm t}/\partial r}\Lambda
	\right)\frac{\partial}{\partial r}\frac{j_\parallel}{B},
$$

in order for the hyperresistive term to be compatible with the ``DiffusionTerm`` implementation. Additionally, since $\sqrt{\langle B^2\rangle}$ is implicitly normalized to $B_{\rm min}$ in the code (not including the factor appearing in the denominator on the LHS, since it is an intrinsic part of the electric field in DREAM), and $j_\Omega/B$ is actually $j_\Omega/(B/B_{\rm min})$, and extra factor $1/B_{\rm min}$ appears in the implementation of the hyperresistive term.


hoppe93:
After further tests and discussions, we have concluded that this modification is not desirable, and the original implementation is in fact more appropriate (see discussion in [Votta *et al* (2026)](https://arxiv.org/abs/2602.22177v1)).

After reviewing the updates, I conclude that there is nothing really we want to salvage from here, so I just close this pull request.

PR #498, opened 2026-02-24
Extended KineticQuantity plotting 

OEmbreus:
Adds new plotting functionality to DREAMOutput.KineticQuantity.

For KineticQuantity.plot() (or, equivalently, DistributionFunction.plot2D()):
* changes interpolateCylindrical to extend pperp to 0 for both negative and positive ppar -- previously only did so for positive ppar.
* adds new argument phaseSpaceWeight: will add the volume element Vp/VpVol * [coordinate jacobian]. For example, in cylindrical, multiplies by the coordinate jacobian pperp / p^2, and uniform in (p, xi) plots. This way, the "area under the graph" will represent the moment of the quantity, allowing the plot to show where the biggest contributions come from.

Add a new plot: KineticQuantity.plotMesh().
Similar to plot(), but uses pcolormesh instead of contourf, with flat shading. This plots on the actual DREAM grid, taking all nodes (p_f, xi_f) and colors the FVM cells by value of the plotted quantities. Can be useful in illustrating/emphasizing grid resolution.

Adds a new resolution monitor calculation: KineticQuantity.resolutionSensors
Returns two KineticQuantities, S_grad and S_curv, representing resolution measures based on first and second-order derivatives, respectively.
* resolutionSensors(axis, quantityWeighted=False): a scale free measure, representing the relative change between adjacent cells in direction "axis" (1: radial, 2: pitch, 3: energy)
* resolutionSensors(axis, quantityWeighted=True): the absolute change between adjacent cells -- will only show where the quantity is large and also varies a lot between cells

Previous behavior:
old_plot2d

New plot with phaseSpaceWeight:
new_plot2d

New plotMesh:
new_plotMesh

S_grad resolution sensor in xi-drection, weighted (dark regions overresolved):
image

These changes necessitated some refactoring of the DREAMOutput.MomentumGrid interface.

p.s. git diff a bit messy because it includes both the py-cleanup and arithmetic branches.


hoppe93:
I fixed most of the Copilot comments, but two remain which maybe @OEmbreus could fix?

OEmbreus:
> I fixed most of the Copilot comments, but two remain which maybe @OEmbreus could fix?

That's great, thanks. I'll get to it during the week.

PR #503, opened 2026-02-28
Optimized PETSc matrix setters and AdvectionDiffusionTerm

OEmbreus:
Introduces a buffered-mode `Matrix::SetElement`, and a new batch column setter `Matrix::SetColumn`. Also introduces many small optimizations to `AdvectionTerm` and `DiffusionTerm`.

I noticed in a test simulation with kinetic electrons and dynamic ions (20 charge states, and using full ion jacobian in the kinetic equation) that ~50% of the total simulation time would sit in `AdvectionDiffusionTerm::SetJacobianBlock`.  With the optimizations in this PR:

- total time inside `AdvectionDiffusionTerm::SetJacobianBlock` reduced by 52%
- total time inside `Matrix::SetElement` reduced by 75%

The net effect is a total simulation time slashed by 31% in this (nearly "worst-case") example. 

This new functionality can likely be incorporated in a few more equation terms, but AdvectionDiffusion (fixed here) is generally the biggest one.

### DIFF
--- a/fvm/Matrix.cpp
+++ b/fvm/Matrix.cpp
@@ -5,6 +5,7 @@
  * 'EquationSystem' class.
  */
 
+#include <algorithm>
 #include <iostream>
 #include <petscmat.h>
 #include "FVM/Matrix.hpp"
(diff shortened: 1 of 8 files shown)


fvm/Matrix.cpp
```diff
@@ -64,7 +64,7 @@ void Matrix::Construct(
     // kept when 'ZeroRows()' is called.
     MatSetOption(this->petsc_mat, MAT_KEEP_NONZERO_PATTERN, PETSC_TRUE);
 
-    // Don't complain about new allocations
+    // Fail if the code attempts to insert a structural nonzero not covered by preallocation.
```
OEmbreus: i think the comment referred to the old commented-out version

OEmbreus:
I realize that the third commit could in principle go into a separate PR. It consists of a set of optimizations that are  independent of the matrix setter modes --- it just cleans up a bunch of computationally intense indexing operations throughout the module. I guess it does fit under the umbrella of "AdvectionDiffusionTerm::SetJacobianBlock optimizations".

PR #514, opened 2026-03-23
Lineintegrated fluid

### LINKED ISSUE #505: Possible bug in FluidQuantity.lineIntegrated()
A possible bug in ``FluidQuantity.lineIntegrated()`` in the Python output interface has been discovered. When comparing against [LIDc](https://github.com/hoppe93/LIDc) and [LineIntegratedDensity](https://github.com/hoppe93/LineIntegratedDensity), ``FluidQuantity.lineIntegrated()`` yields a factor ~2.5 larger values, while the two other tools are in good agreement. All three tools use different algorithms, and so can be considered as three different implementations, although LineIntegratedDensity and ``FluidQuantity.lineIntegrated()`` should use the most similar approaches.

I propose to set up a verification test in which the analytic B module of DREAM is used to generate first a circular plasma and then, if needed, including various shaping parameters. The electron density can be set to a constant value, which should allow for straightforward analytical integration of the density.

ekmanlinn:
The PR contains:

- Changing of the name of the old `lineIntegrated()` to `lineIntegrated_SPI` since it calculates the line integration from a "SPI" coordinate system. Remember this still don't work for all equilibrium, but maybe this isn't a problem?
- Implemented a new `lineIntegrated()` based on the same algorithm as in `[LIDc](https://github.com/hoppe93/LIDc)`, maintaining the speed of doing such integrals.

Not concerning this issue but since it was such a small fix, I also added
- Documentation for the VOLUMETRIC ion source implementation
- The smallest fix in the NBI documentation 

Fixes #505


hoppe93 (changes requested):
Thanks, this looks good! I just have one comment in the documentation :)

doc/sphinx/source/frontend/api/Settings/eqsys/ions.rst
```diff
@@ -589,6 +589,60 @@ Finally, particles in **several charge states** may be added simultaneously:
    Thus, the unit of ``dNdt`` is simply :math:`\mathrm{s}^{-1}` (or particles
    per second).
 
+Radial Injection Example
+^^^^^^^
```
hoppe93, line 593: This should also be
```
Radial Injection Example
^^^^^^^^^^^^^^^^^^^^^^^^
```

hoppe93 (changes requested):
Still one comment remaining...

hoppe93 (approved):
Great!

PR #515, opened 2026-03-28
PR1: package-backed stellarator geometry foundation

rogeriojorge:
## High-level summary
This PR is the first infrastructure layer in a stacked stellarator geometry series.

The current `stellarator` branch already has an important strength: DREAM can already run reduced stellarator cases inside the existing DREAM time-dependent framework, so users can benefit from the same mature solver stack, current evolution machinery, electric-field evolution, prescribed/ self-consistent equation infrastructure, and Python settings interface that already make DREAM productive on the tokamak side.

This PR builds on that strength. It does **not** try to rewrite the stellarator physics model or change the C++ kernel equations. Instead, it focuses on making stellarator geometry **easier to ingest, cache, test, reload, compare, and reuse** from Python.

In practice, this PR introduces a package-backed geometry workflow so that VMEC/DESC-derived geometry can be treated as a first-class, reproducible object in the DREAM frontend.

## What already works well on the stellarator branch
Before this PR, the branch already provided several things that are worth preserving and building on:

- DREAM already had a stellarator-capable radial-grid path and could evolve reduced stellarator cases inside the standard DREAM framework.
- The branch already benefited from DREAM's solver maturity: nonlinear solves, time stepping, current evolution, electric-field evolution, ion handling, and diagnostic/output infrastructure were already there.
- The branch already had the core idea of using flux-surface-reduced stellarator geometry inside DREAM, which is the right level of abstraction for a first integrated stellarator workflow in this codebase.
- The branch already exposed stellarator functionality through the Python interface, which makes it realistic to build practical equilibrium-loading and regression-test workflows around it.

This PR is deliberately designed to preserve those advantages while making the geometry layer more modular and much easier to validate.

## What this PR adds
This PR adds the first package/provider layer for stellarator geometry.

The main additions are:

- A versioned `StellaratorGeometryPackage` HDF5 contract in `py/DREAM/Settings/StellaratorGeometry.py`.
- A provider abstraction in `py/DREAM/Settings/StellaratorGeometryProviders.py`.
- Frontend integration in `py/DREAM/Settings/StellaratorMagneticField.py` and `py/DREAM/Settings/RadialGrid.py`.
- A Python-side flux-tube evaluator in `py/DREAM/Settings/StellaratorFluxTube.py`.
- Reproducible examples in `examples/stellarator/`.
- A dedicated stellarator geometry smoke/regression suite in `tests/physics/stellarator_geometry/stellarator_geometry.py`.

### 1. Versioned geometry packages
The new `StellaratorGeometryPackage` is the canonical geometry container introduced in this series.

It stores:
- metadata such as provider, source kind, radii, and field-period count,
- 1D profiles such as `iota`, `G`, `I`, `psi_T`, `B_min`, `B_max`, and `f_passing`,
- the sampled arrays the current DREAM kernel/frontend path already uses,
- an optional spectral/Boozer block when available.

Why this matters:
- geometry can now be written once and reloaded reproducibly,
- provider output becomes testable independent of runtime imports,
- later PRs can compare package-backed and provider-backed workflows directly.

### 2. Provider-based geometry loading
This PR introduces a `GeometryProvider` abstraction and the first concrete providers:
- `DescProvider`
- `VmecJaxProvider`
- `PackageProvider`

Why this matters:
- equilibrium-source logic is separated from frontend/kernel hydration,
- VMEC/DESC/package workflows share one frontend path,
- later PRs can add richer source-specific logic without reworking `RadialGrid` again.

### 3. Package-backed frontend hydration
`RadialGrid.setStellarator(...)` can now load stellarator geometry through a provider and hydrate the same frontend state from a package object.

Why this matters:
- frontend state no longer depends only on ad hoc numeric caches,
- roundtrips are testable,
- cache reuse becomes much easier to reason about.

### 4. Flux-tube evaluation hooks
This PR adds a Python-side `FluxTubeEvaluator`.

Why this matters:
- field-line / flux-tube diagnostics become available without touching the C++ kernel,
- later PRs can enrich the spectral/Boozer content while keeping a stable Python diagnostic entry point,
- reviewers can inspect geometry in a way that is much more intuitive than raw flattened arrays.

### 5. Reproducible examples and QA data
This PR adds a small QA VMEC fixture and examples that exercise the new workflow:
- `geometry_from_package.py`
- `geometry_from_vmec.py`
- `geometry_from_desc.py`
- `package_no_bootstrap_smoke.py`

Why this matters:
- the branch gains a concrete, inspectable geometry workflow,
- users can see input/output behavior directly,
- the test suite can exercise the new path on a real equilibrium.

### 6. Dedicated smoke and regression coverage
This PR adds a focused stellarator geometry test slice and wires it into the physics test runner.

Why this matters:
- package roundtrip behavior is tested,
- settings roundtrips are tested,
- package-backed smoke runs are tested,
- the new layer becomes a stable base for later PRs.

### Canonical package first, provider second
The most important decision in this PR is to make the **package** the canonical intermediate representation.

This is intentional.

If DREAM is going to support multiple stellarator equilibrium sources cleanly, the code should not let each source define its own frontend behavior. The provider should build a package; the frontend should consume the package.

That decision makes several later things easier:
- caching,
- regression testing,
- cross-provider comparisons,
- package-backed no-bootstrap smoke cases,
- future geometry tooling that does not depend on a live equilibrium object being importable.

### No physics rewrite in PR1
This PR does **not** change the stellarator physics model solved by the kernel.

That is also intentional.

The first step should be to make geometry ingestion and validation robust before changing bootstrap closures, wall coupling, or deeper stellarator-specific self-consistent physics. Otherwise it becomes hard to tell whether a regression is geometric, frontend-related, or truly physical.

### Flux-tube support stays in Python here
The initial flux-tube evaluator is Python-side.

That gives two benefits:
- geometry can be inspected immediately,
- we avoid expanding the kernel review surface before the geometry contract has stabilized.

## What this PR does **not** try to do
This PR is intentionally limited.

It does not:
- change the underlying stellarator kernel equations,
- introduce a new bootstrap-current model,
- rewrite wall/transformer/current-closure logic,
- make the C++ kernel consume Boozer/spectral geometry directly.

Those are important follow-up topics, but they should come after the geometry/package layer is in place and tested.

## Suggested figures / diagrams for the PR description
These do not need to live in the repository. They are meant only as helpful PR visuals.

### Diagram A: Geometry workflow overview
```mermaid
flowchart LR
    A["VMEC wout / DESC equilibrium / package file"] --> B["GeometryProvider"]
    B --> C["StellaratorGeometryPackage"]
    C --> D["RadialGrid.setStellarator(...)"]
    C --> E["FluxTubeEvaluator"]
    D --> F["Package-backed DREAM smoke case"]
```

### Figure B: Fixed-phi geometry cross-section
Suggested content:
- plot several flux surfaces at one toroidal angle,
- compare provider-built geometry with package-reloaded geometry,
- caption idea: "Package-backed reload reproduces the same geometry cross-section as the provider-built frontend state."

### Figure C: Example flux-tube trace
Suggested content:
- one sample field-line/flux-tube trace,
- show `R`, `Z`, or `|B|` along the line,
- caption idea: "The new Python-side flux-tube evaluator gives an immediate diagnostic view of the packaged geometry."

### Figure D: 1D profile summary
Suggested content:
- plot `iota`, `B_min`, `B_max`, and `f_passing` versus radius,
- caption idea: "The package contains both the sampled geometry used by DREAM and compact 1D profiles useful for validation and comparison."

### Figure E: Package workflow / reproducibility figure
Suggested content:
- one panel showing: source equilibrium -> package file -> reload -> no-bootstrap smoke run,
- caption idea: "This PR makes stellarator geometry reloadable and testable as a first-class artifact."

## Validation performed in this PR
- `python tests/physics/runtests.py stellarator_geometry --verbose`
- `python examples/stellarator/geometry_from_package.py`
- `python examples/stellarator/geometry_from_vmec.py`
- `python examples/stellarator/geometry_from_desc.py` when the optional DESC stack is available
- `python examples/stellarator/package_no_bootstrap_smoke.py --provider auto --source examples/stellarator/data/stellarator_geometry_v1.h5`

## Follow-up PRs in the stack
This PR is intentionally the foundation layer. The follow-up stack is:

- **PR2**: improve Boozer loading and add package-vs-legacy kernel parity checks.
- **PR3**: add DESC-native Boozer and flux-tube support.
- **PR4**: clean up the provider-facing API and make the workflow friendlier/documented.
- **PR5**: add a maintained end-to-end stellarator workflow example.
- **PR6**: clearly separate maintained examples from legacy exploratory scripts.
- **PR7**: add generated-package parity gates for provider-vs-package frontend loads.
- **PR8**: add a user-facing geometry comparison utility.

So the intent is:
1. make geometry a stable artifact,
2. strengthen provider/spectral support,
3. improve user workflow and reviewability,
4. add stronger parity gates and diagnostics.

## Why I think this is a good first PR
This PR gives the stellarator branch something it did not have before: a clean, testable, package-backed geometry layer that is useful immediately, while staying conservative about solver physics.

That is the right tradeoff for the first PR in the series.

It keeps what is already valuable in DREAM's stellarator branch, adds a practical workflow that users can run today, and creates a much better base for the more ambitious stellarator-specific follow-up work.


rogeriojorge:
This PR is still a draft; there is no need to review it yet.

rogeriojorge:
Superseded by the fork-based stellarator PR stack in rogeriojorge/DREAM.

PR #520, opened 2026-05-06
Fixed two bugs in kinetic Compton source

IdaEkmark:
I removed the double initialization of `id_ntot` and changed the order of the jacobian if statements so that the derivation with regards to $n_{\rm tot}$ in fluid mode would be correct


ekmanlinn:
So n_tot shouldn't be calculated when Fluid mode is on? Isn't that what happens now ? If fluid mode is on but derivId = n_tot then it will return 0. Maybe this is the intention, just wanted to double check.

IdaEkmark:
Not really - this regards the differentiation of the Compton source term in the Jacobian. If we run in fluid mode, we will evaluate the derivative w.r.t. $n_{\rm tot}$ on line 203 (previously 206) that is `else if(derivId==id_ntot) {`, which will be different from if we run in kinetic mode.

ekmanlinn (approved):
Looks good :)

PR #521, opened 2026-05-12
Fixed tau_coll bug

IdaEkmark:
If you have start a simulation from another output and in the first output have hot-tail enabled but in the second output turn it off, you get the error that `tau_coll` does not exist in the equation system (which should not be a problem if the fluid hottail model is not utilized). I have solved this in this pull request.


hoppe93 (approved):
Looks good to me!

PR #523, opened 2026-05-25
Add support for initializing j_hot

hoppe93:
In isotropic simulations, we currently have to run an initialization simulation with only one purpose: setting a desired initial value of ``j_hot``. It should be possible to accomplish this much more easily, and so in this pull request I propose a more user-friendly approach for initializing the isotropic model.


hoppe93:
Currently there is still some problem with the way the simulations are set up, so that they still fail when initializing ``j_hot`` in this way. More work is required...

PR #526, opened 2026-06-10
Fix typos in tritium sections in n_re and ions in docs

BjornZaar:
Found a few typos regarding the tritium seed in the docs. The half-life was incorrect in "ions", and the code example in "n_re" would not run. In addition, I added a few ```\rm``` in the equations in the Compton section where it was missing


hoppe93 (approved):
Looks great, thanks for submitting! :D

PR #527, opened 2026-06-17
Added logos with transparent background

IdaEkmark:
In the original one, there is an "egg" to achieve the half moon effect. I've removed the necessity of the egg


hoppe93 (approved):
Excellent!

PR #529, opened 2026-06-23
Fixed bug for restarting simulation with scrape off

IdaEkmark:
If you initialize a simulation from another with scrape off, it crashed because the scrape off class does not take a momentumgrid as an input, similar to OtherIonSpeciesFluidQuantity. I therefore did the same solution for LCFSLoss as for OtherIonSpeciesFluidQuantity

### DIFF
--- a/py/DREAM/Output/LCFSLoss.py
+++ b/py/DREAM/Output/LCFSLoss.py
@@ -6,7 +6,7 @@ from .ScalarQuantity import ScalarQuantity
 
 class LCFSLoss(OtherFluidQuantity):
 
-    def __init__(self, name, data, description, grid, output):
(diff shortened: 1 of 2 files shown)


hoppe93:
I think the fix included in ``trigger-isotropic`` would be preferred, as it doesn't require adding a special case. It might still take a while to merge that branch, so having a fix for this problem before it gets merged is probably nice. But could we implement in the same way?

https://github.com/chalmersplasmatheory/DREAM/pull/518/changes#diff-f22c294b1606c17cb311ec857e17bedd86812e69acc08a82aa62641ca5e613fd

hoppe93 (approved):
Looks good!

PR #532, opened 2026-06-26
Fix animation bug with matplotlib 3.11

### LINKED ISSUE #531: Animations no longer work in Python 3.14 + Matplotlib 3.11
Recent updates to Matplotlib means that the way DREAM calls matplotlib to generate animations no longer works. The problem seems to be that DREAM passes integers of type ``numpy.int64`` to matplotlib, which causes an error.

hoppe93:
Fixes the bug reported in #531 which appears in matplotlib 3.11 and which causes ``FluidQuantity.animate()`` and ``KineticQuantity.animate()`` to no longer work.


IdaEkmark (approved):
Looks good!

PR #533, opened 2026-06-29
Compton tritium bug

IdaEkmark:
The classes to evaluate densities of electrons to remove to compensate for kinetic Compton and Tritium where derived from the wrong class, meaning they gave the wrong numbers. I have fixed this, and also moved out the classes from the f_hot-file for a cleaner code

### DIFF
--- a/src/Equations/Fluid/TotalElectronDensityFromKinetic/TotalElectronDensityFromKineticCompton.cpp
+++ b/src/Equations/Fluid/TotalElectronDensityFromKinetic/TotalElectronDensityFromKineticCompton.cpp
@@ -1,0 +1,31 @@
+#include "DREAM/Equations/Fluid/TotalElectronDensityFromKinetic/TotalElectronDensityFromKineticCompton.hpp"
+
+/**
+ * Implementation of an equation term which represents the total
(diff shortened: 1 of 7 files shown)


src/Equations/Fluid/TotalElectronDensityFromKinetic/TotalElectronDensityFromKineticCompton.cpp
```diff
@@ -0,0 +1,42 @@
...
+{
+    this->limit = 1000;
+    this->wp = gsl_integration_workspace_alloc(limit);
+    this->wpOut = gsl_integration_workspace_alloc(limit);
+}
+
+TotalElectronDensityFromKineticCompton::~TotalElectronDensityFromKineticCompton() {
+    if (wp)
+        gsl_integration_workspace_free(wp);
+    if (wpOut)
```
hoppe93: I think it would be best to explicitly compare ``wp`` and ``wpOut`` to ``nullptr`` to prevent C++ compiler curiosities.

hoppe93 (changes requested):
Mostly looks very good! I just have one comment.

hoppe93 (approved):
Looks great now!

PR #540, opened 2026-07-21
Clamp BOverBmin

leferi99:
Clamp BOverBmin to avoid issues with numerical geometry where one can get B/Bmin ~ (1 - epsilon), making the sqrt() argument negative, and causing a crash.


leferi99:
I made the edit at the wrong location in the file...

PR #541, opened 2026-07-21
Clamp BOverBmin value in KnockOnUtilities.cpp

leferi99:
Clamp BOverBmin to avoid issues with numerical geometry where one can get B/Bmin ~ (1 - epsilon), making the sqrt() argument negative, and causing a crash.

Now the fix is at the correct place:
```cxx
    BOverBmin = B_ / Bmin;
    // Clamp BOverBmin to avoid issues with numerical geometry
    BOverBmin = std::max(BOverBmin, 1.0);
    xi0Cutoff = sqrt(1 - 1 / BOverBmin);
```

### DIFF
--- a/src/Equations/KnockOnUtilities.cpp
+++ b/src/Equations/KnockOnUtilities.cpp
@@ -213,6 +213,8 @@ bool KnockOnUtilities::CheckIfReachableAndSetGeometricQuantities(
         ir, theta, B_, Jacobian, ROverR0_, NablaR2_, FVM::FLUXGRIDTYPE_DISTRIBUTION
     );
     BOverBmin = B_ / Bmin;
+    // Clamp BOverBmin to avoid issues with numerical geometry
+    std::max(BOverBmin, 1.0);    
     xi0Cutoff = sqrt(1 - 1 / BOverBmin);
 
     if (xi0Cutoff >= fabs(xi01)) {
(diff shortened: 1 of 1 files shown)


leferi99:
I should also say that I tested it, and after I fixed the obvious oversight of no value allocation, DREAM doesn't crash with this approach.

src/Equations/KnockOnUtilities.cpp
```diff
@@ -213,6 +213,8 @@ bool KnockOnUtilities::CheckIfReachableAndSetGeometricQuantities(
         ir, theta, B_, Jacobian, ROverR0_, NablaR2_, FVM::FLUXGRIDTYPE_DISTRIBUTION
     );
     BOverBmin = B_ / Bmin;
+    // Clamp BOverBmin to avoid issues with numerical geometry
+    BOverBmin = std::max(BOverBmin, 1.0);
```
IdaEkmark, line 217: Is it a good idea to have it on one line?
`BOverBmin = std::max(B_ / Bmin, 1.0);

IdaEkmark (commented):
I see no problems with this implementation, just had a suggestion

src/Equations/KnockOnUtilities.cpp
leferi99: I would think, generally, it's better for tracking changes if there isn't too much going on in one line, but given this is a minor thing and it has a comment, it shouldn't really matter here, in my opinion (but I am quite far from an expert in coding).

Issue #545, opened 2026-08-03
Optimizations on Intel compilers do not allow infinites by default


hoppe93:
When compiling DREAM using an Intel compiler, it has been found that the compiler assumes that infinite numbers cannot appear. This causes problems in some parts of the code which rely on infinites for certain computations.

PR #546, opened 2026-08-03
Add -fno-finite-math-only flag to Intel compiler

hoppe93:
This pull request solves issue #545 by adding the flag ``-fno-finite-math-only`` in release mode to the Intel compiler. The assumption of "finite math only", which seems to be done at least in some Intel compilers when applying aggressive optimizations (but not GCC or Clang), violates the IEEE 754 standard on floating-point arithmetic, and (more importantly) assumptions made in certain parts of DREAM that some operations lead to infinites.

This fix should therefore maintain strong optimizations in the Intel compiler, but avoid the assumption that no infinite numbers can arise.


hoppe93:
Regarding the suppressed comment from Copilot: this issue only concerns the Intel oneAPI compiler, not the IntelLLVM compiler.

IdaEkmark (approved):
Looks good!

PR #549, opened 2026-08-26
Change trapz -> trapezoid

hoppe93:
Add support for the newer ``numpy.trapezoid`` function. This function was added in numpy 2.0.0, while ``trapz`` was at the same time deprecated. The ``trapz`` function was ultimately removed in numpy 2.4.0. This fix ensures that trapezoidal integration works across all versions of numpy.


ekmanlinn (approved):
Looks good!

PR #550, opened 2026-09-08
Fix T_cold.plotEnergyBalance(log=True)

hoppe93:
Previously, the log argument was not passed to plotIntegral(), so that the plot did not render as expected. This pull request fixes this bug.


ekmanlinn (approved):
Looks good! :)

PR #551, opened 2026-09-11
Fixed bug that resulted in seg fault for Kiramov BC

IdaEkmark:
When I ran a simulation with the Kiramov BC on master, I had a segmentation fault due to how the transportOperator-object was deleted. This fixed it


hoppe93 (approved):
Looks great!

